PR #51, opened 2017-01-07
Safeguard against wrong Beam=BeamContinue on quarters and longer notes

th-we:
This commit solves problems with the following file:

[single-note-beam-bug.sib.zip](https://github.com/music-encoding/sibmei/files/691580/single-note-beam-bug.sib.zip)

Is there a coding style to follow when contributing?  This commit introduces a pretty long line and I wonder whether and how to wrap it.


ahankinson:
Looks good. Thanks!

PR #52, opened 2017-01-08
Change logic for detecting tuplet ends

### LINKED ISSUE #45: One tuplet encoded as multiple <tuplet> elements
Some tuplets in the Sibelius file are encoded as multiple <tuplet> elements in the mei file, instead of just one <tuplet> element.

When the base of a tuplet is ‘2’, if its numerator is greater than 3; the tuplet is encoded as more than one <tuplet> element in the mei file. Compare measures 4 and 6 to measure 2 in “tuplets_1”.

I think that this can be the cause that certain tuplets even get merged together sometimes, as in measure 8 of “tuplets_1” and measures 1 and 2 of “tuplets_2”; in the latter, this merge of tuplets even change the order of the notes.

[tuplet_issues.zip](https://github.com/music-encoding/sibmei/files/410020/tuplet_issues.zip)

th-we:
I don't really understand `tuplet.Duration` as it has pretty odd values, but it obviously is the cause of the duplicate tuplet issue. I tried different logic which seems to work, but this pull request needs some pair of experienced eyes to check whether it's actually sound.

Fixes #45

### DIFF
--- a/src/Utilities.mss
+++ b/src/Utilities.mss
@@ -235,14 +235,17 @@ function TupletsEqual (t, t2) {
 
 function IsLastNoteInTuplet (bobj) {
     //$module(Utilities.mss)
-    if (bobj.ParentTupletIfAny = null or bobj.GraceNote = True)
+    next_obj = bobj.NextItem(bobj.VoiceNumber, 'NoteRest');
+
+    if (next_obj = null)
     {
-        return false;
+        return true;
     }
 
     tuplet = bobj.ParentTupletIfAny;
+    next_obj_tuplet = next_obj.ParentTupletIfAny;
 
-    if (bobj.PositionInTuplet >= tuplet.Duration)
+    if (next_obj_tuplet = null or tuplet.Position != next_obj_tuplet.Position)
     {
         return true;
     }


ahankinson:
Can you test files to show the before and after behaviour? A sibelius file, the before and after MEI would be great.

ahankinson:
Also, you should be submitting these against `develop,` not master....

ahankinson:
I've just corrected the base branch.

th-we:
Sorry - I was lazy because there are examples in the linked issue, but certainly no before/after comparison. Here's a comparison:

[tuplet_example.ZIP](https://github.com/music-encoding/sibmei/files/692199/tuplet_example.ZIP)

ahankinson:
Brilliant! I just tested it on the MEI test set we have and it works perfectly. Thank you very much. (I know how tricky the tuplet code can be, and I'm glad it turned out to be a relatively simple fix).

ahankinson:
Updated test set here:

http://ddmal.github.io/mei-test-set/MEI-test-set/tuplets.html

th-we:
#45 should be closed now. What's the policy? May I go ahead and do that?

PR #57, opened 2017-01-31
Add @metcon="false" to measures not conforming to time signature

th-we:
This is to make it easier for consuming applications to cope with pickup bars and the like.

I'm not sure what's the best way to access the current time signature - I passed it as argument, but if that's undesired, please let me know.

In contrast to the `ExternalBarNumberString` check, I put this outside the loop through all staffs.

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -301,6 +301,15 @@ function GenerateMeasure (num) {
     // since so much metadata about the staff and other context
     // is available on the bar that should now be on the measure, go through the bars
     // and try to extract it.
+    
+    systf = score.SystemStaff;
+    currTimeS = systf.CurrentTimeSignature(num);
+    sysBar = systf[num];
+    
+    if (sysBar.Length != (currTimeS.Numerator * 1024 / currTimeS.Denominator))
+    {
+        libmei.AddAttribute(m, 'metcon', 'false');
+    }
 
     for i = 1 to staves.Length + 1
     {


ahankinson:
Are you basing this calculation only on the first staff? (`NthStaff(1)`)

I've been trying to keep function arguments as few as possible, preferring instead to figure out context within the method itself if at all possible. If there's no way to get the current time signature inside `GenerateMeasure`, I would prefer to do it one level up ('GenerateMEIMusic').

th-we:
I saw that `@label` is set during the loop over all staves - `@metcon` could be set there as well. I'd however expect that the attribute will be re-set with the same value for each iteration, wouldn't it? I don't think different staffs could have different durations, unless there's something wrong with Sibelius.  Not sure whether `ExternalBarNumberString` can differ from staff to staff - I'm a total ManuScript novice.

On the second topic: Then what about introducing `Self._property:CurrentTimeSig` in `GenerateMEIMusic`?

ahankinson:
Yeah, good point -- I could do a check in there to see if the measure already has `@label` set (using `libmei.GetAttribute(m, 'label')`, which will return `False` if it doesn't exist). 

The underlying problem is that Sibelius has an inverted hierarchy than MEI (Staff -> Measure instead of Measure -> Staff) so to get to the measure level you first need to go through the staves. 

Doing a bit of digging it appears Sibelius doesn't (really) support polyrhythms, since the way to access the current time signature is always through the "SystemStaff" object. Still, I think it's better to get the time signature from the system staff, rather than the first actual staff. You can do this anywhere in the code:

```
score = Self._property:ActiveScore;
systf = score.SystemStaff;
currTimeS = systf.CurrentTimeSignature(barnum);
```

Another thing to watch out for: I would avoid chaining indexes from return values, like `score.NthStaff(1)[num]`. Sometimes it works, sometimes it doesn't. Instead, you should break this into two statements, since that always works. It's annoying, but I've been bitten by problems with that in the past.

th-we:
Many thanks for sharing your insights, I highly appreciate that. So what about the updated pull request?

I didn't try moving the `ExternalBarNumberString` out of the loop as I have no idea how to actually create such a bar number string thing in Sibelius to test it. Re-setting the attribute doesn't seem like such a big deal.

ahankinson:
Looks better! Maybe you could attach a sample Sibelius file? I could add it to the DDMAL/mei-test-set to make sure it goes into our test collection.

th-we:
The test file for the flawed ContinueBeam works for this as well.

[continuebeam-after-rest.sib.zip](https://github.com/music-encoding/sibmei/files/747092/continuebeam-after-rest.sib.zip)

ahankinson:
Sorry -- could you include the Sibelius file as well? That way we can keep testing the conversion process, instead of just the result.

th-we:
I'm still on Sibelius 8.1.

ahankinson:
Hopefully just one more request -- could you send it along as a Sibelius 7.5 file please?

th-we:
Here's 7.5:

[continuebeam-after-rest-Sib7.5.sib.zip](https://github.com/music-encoding/sibmei/files/750826/continuebeam-after-rest-Sib7.5.sib.zip)

PR #60, opened 2017-02-03
score.MusicTextFontName is for dynamics, not lyrics

th-we:
Looks funny if "Opus Text Std" (apparently the default for `score.MusicTextFontName`) is used in Lyrics.


ahankinson:
Looks reasonable. I won't test it, but will trust that you're right! :)

PR #61, opened 2017-02-03
Avoid occasional error "ProcessSymbol:12:Field bar not found"

th-we:
I occasionally get the error `ProcessSymbol:12:Field bar not found` - not sure when and why exactly. But this seems to fix it.

To avoid the additional parameter, we could also use `Self._property:CurrentBar`. Would that be preferable?

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -638,7 +638,7 @@ function GenerateLayers (staffnum, measurenum) {
 
     for each SymbolItem sobj in bar
     {
-        ProcessSymbol(sobj);
+        ProcessSymbol(sobj, bar);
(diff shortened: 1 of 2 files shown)


ahankinson:
Thanks. I just committed an alternate fix for this problem to the develop branch. Since symbol objects are derived from `BarObject` they always have the `ParentBar` property, so it's easy enough to get that from the existing `sobj` parameter.

ahankinson:
If it fixes your problem, please feel free to close this PR. I'll keep it open until then.

th-we:
Just for reference, the alternative commit is b995e6a. I'm sure it fixes the problem.

PR #65, opened 2017-02-04
Fix crahs for empty measure with clef

th-we:
The attached test file has cross staff notation where one staff has no content at all, except for a clef at the end of the measure. This currently crashes sibmei as it tries to attach the clef to the last existing layer, but no layer has been created yet.

This patch creates a layer for clefs that occur before anything else in the measure. If the clef is not at position 0, it also creates a `<space>` element to fill the empty room in the layer.

[empty-measure-with-clef.sib.zip](https://github.com/music-encoding/sibmei/files/752421/empty-measure-with-clef.sib.zip)


th-we:
There's a more general problem behind this that should be fixed properly, so I withdraw this pull request until I have a proper fix.

PR #66, opened 2017-02-04
Support fermatas on BarRests

th-we:
I re-used the fermata generation code for `BarRest`s. To avoid identical copy&paste code in 6 places, I created `GenerateFermata(bobj, anchor, shape)`.

If there's problem with the multiple arguments, please let me know.  What's the reason for preferring single-argument functions? Are there ManuScript specific problems with more than one argument or is there a general coding philosophy behind this?

[fermatas.sib.zip](https://github.com/music-encoding/sibmei/files/752574/fermatas.sib.zip)

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -779,43 +779,15 @@ function GenerateNoteRest (bobj, layer) {
         libmei.AddAttribute(nr, 'grace', 'acc');
     }
 
-    if (bobj.GetArticulation(PauseArtic))
+    if (bobj.GetArticulation(PauseArtic) or bobj.GetArticulation(TriPauseArtic) or bobj.GetArticulation(SquarePauseArtic))
     {
-        fermata = libmei.Fermata();
-        libmei.AddAttribute(fermata, 'form', 'norm');
-        libmei.AddAttribute(fermata, 'shape', 'curved');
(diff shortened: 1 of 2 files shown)


th-we:
Updated the code.

I know there are not many active contributors, but would it be worth to compile some coding guidelines?

ahankinson:
Thank you!

ahankinson:
What sort of guidelines are you looking for?

ahankinson:
Looking at it this seems to produce the following MEI:

```
<fermata xml:id="m-37" form="norm" layer="1" shape="curved" staff="1" startid="#m-36" tstamp="1"/>
```

I think the mix of startid and tstamp are confusing Verovio:

verovio

Would it be ok if I removed the call to `AddBarInfoToElement` (L1482)?

lpugin:
Verovio will use startid and ignore tstamp when both are given.

ahankinson:
But it looks like it's drawing the fermata on the `mRest` at tstamp=1.

lpugin:
Ah yes, this is a bug. I will still better to use startid because the tstamp of the middle of the measure (e.g., 2.0 in 3/4) will not be necessary in the middle of the measure depending on what is happening at other voices.

ahankinson:
Yes, of course. My suggestion for this issue is to remove the `tstamp` attribute.

PR #67, opened 2017-02-16
Beam rewrite: Support grace beams/nested beams and @breaksec

th-we:
This is a complete rewrite of `ProcessBeam()`, but I think that function is much clearer now and works for the attached test cases:

[nested-beams.sib.zip](https://github.com/music-encoding/sibmei/files/781837/nested-beams.sib.zip)


ahankinson:
Thank you!

PR #69, opened 2017-02-18
Support nested tuplets with linked list

th-we:
This is a rewrite of `ProcessTuplet()` and related code to support nested tuplets. I've implemented two variants for keeping track of and comparing tuplet nesting between notes. One uses a linked list (this pull request), the other one (#68) an array representing a tuplet stack. Which method is preferable?

[nested-tuplets.sib.zip](https://github.com/music-encoding/sibmei/files/784838/nested-tuplets.sib.zip)

### DIFF
--- a/src/Utilities.mss
+++ b/src/Utilities.mss
@@ -234,25 +234,84 @@ function TupletsEqual (t, t2) {
 
 }  //$end
 
-function IsLastNoteInTuplet (bobj) {
+function CountTupletsEndingAtNoteRest(noteRest) {
     //$module(Utilities.mss)
-    next_obj = bobj.NextItem(bobj.VoiceNumber, 'NoteRest');
-
-    if (next_obj = null)
+    tuplet = noteRest.ParentTupletIfAny;
+  
+    if (tuplet = null)
     {
-        return true;
+        return 0;
     }
+  
+    nextNoteRest = noteRest.NextItem(noteRest.VoiceNumber, 'NoteRest');
 
-    tuplet = bobj.ParentTupletIfAny;
-    next_obj_tuplet = next_obj.ParentTupletIfAny;
-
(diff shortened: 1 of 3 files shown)


src/Utilities.mss
```diff
@@ -234,25 +234,84 @@ function TupletsEqual (t, t2) {
...
     //$module(Utilities.mss)
-    next_obj = bobj.NextItem(bobj.VoiceNumber, 'NoteRest');
-
-    if (next_obj = null)
+    tuplet = noteRest.ParentTupletIfAny;
+  
+    if (tuplet = null)
     {
-        return true;
+        return 0;
```
ahankinson: What is this return value? I would prefer returning 'false' or 'null' if you mean to not return anything, or to break out of the function.

src/Utilities.mss
```diff
@@ -234,25 +234,84 @@ function TupletsEqual (t, t2) {
...
-    if (next_obj_tuplet = null or tuplet.Position != next_obj_tuplet.Position)
+    if (nextNoteRest = null or nextNoteRest.ParentTupletIfAny = null)
     {
-        return true;
+        return GetSibTupletDepth(noteRest);
     }
+    
+    if (TupletsEqual(tuplet, nextNoteRest.ParentTupletIfAny))
+    {
+        return 0;
```
ahankinson: Same here.

src/Utilities.mss
th-we: This function counts how many tuplets end at a specific `NoteRest`.  If the `NoteRest` is not part of a tuplet, no (i.e. 0) tuplets end here.

th-we:
The array approach is indeed simpler and more maintainable. I wasn't sure how ManuScript would handle the creation and garbage collection of that many arrays. But I guess I was over-wary. It seems to be handling it just fine. So scrap the linked list approach.

PR #70, opened 2017-02-25
Check for existance needs "!= null" test

th-we:
One of those nasty things that one is used to from JavaScript et al. which works differently in ManuScript bit me here...

### DIFF
--- a/src/Utilities.mss
+++ b/src/Utilities.mss
@@ -287,7 +287,7 @@ function GetMeiTupletDepth (layer) {
     //$module(Utilities.mss)
     depth = 0;
     tuplet = layer._property:ActiveMeiTuplet;
-    while (tuplet)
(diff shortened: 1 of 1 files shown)


ahankinson:
Thanks!

ahankinson:
How much of a problem is this? Should I release 2.1.1 to fix it?

th-we:
Sometimes a non-null/non-false value evaluates to false in ManuScript, right? This means, for a nested tuplet, it might exit the loop prematurely.  That was however not a problem in the test cases - not sure whether it *can* be a problem for other cases.

People might run into problems when they have nested tuplets, which is pretty rare I guess.  So for my part, I'm O.K. with keeping calm and putting it in the next release.

ahankinson:
OK.

PR #72, opened 2017-03-06
Support clefs inside of tuplets and beams

th-we:
Previously, clef changes inside tuplets or beams were moved outside of the beam or tuplet.

Test file:

[clefs-in-tuplets-and-beams.sib.zip](https://github.com/music-encoding/sibmei/files/821590/clefs-in-tuplets-and-beams.sib.zip)


src/Utilities.mss
```diff
@@ -462,3 +474,75 @@ function GetNongraceParentBeam (noteRest, layer) {
     return null;
 }  //$end
 
+function AppendToLayer (meielement, l, beam, tuplet) {
```
th-we: The motivation for this function is to properly append everything that can be inside a beam or tuplet.  However it can be used to append anything to a layer.

The elements it's currently used for are `<note>`s, `<rest>`s, `<chord>`s, `<space>`s and `<clef>`s.  It should be used for `<fTrem>`s and `<bTrem>`s as well when support for them is added.

As far as I can see, the only elements that do not rely on this function are `<mRest>`, `<mRpt>` and `<mRpt2>`.  They don't need the tuplet and beam checking overhead but *could* no problem be appended using this function.  For consistency's sake, that might be worth a consideration.

rettinghaus:
@th-we What's the status with this? Has it been addressed in another PR or could you resolve those conflicts?

th-we:
Phew, was a challenge, but fortunately the conflicts could still be resolved.

rettinghaus:
@annplaksin Clefs do need `@staff`, but I'm not sure either about `@tstamp`. As other layer elements get a `@tstamp.ges`, this would be better for consistency. 
Or should we drop `@tstamp` / `@tstamp.ges` support for layer elements completely?

th-we:
> Looks good....
> But I have just one question concerning the output for the clef as a first element in a bar: Why does it have `@tstamp="3"`?

You're talking about this, right?

image

The spacing doesn't really visualize the data well.  It's actually a 4/4 bar where the music starts in the second half.  Typical cases where something like this can be found is in piano music where a voice continues on the other staff:

image

Inputters might be "disturbed" by the grayed out hidden rests and delete them altogether. This is not good practice, but it can be found in the wild and we should make sure that it does not crash Sibmei.

If it helps to make the intention of the test clearer, I can turn the test file into a grand staff and use the situation illustrated above instead of the strange bar.

> 
> But one question before we merge: Do we need `@staff` and `@tstamp` for clefs?

Oh, there are a lot of attributes that we don't really need, I think.  (Do we need `@tstamp.real` for example? Not sure for what reason that was added.)  Any specific reason to remove `@staff` and `@tstamp`?

annplaksin:
> The spacing doesn't really visualize the data well. It's actually a 4/4 bar where the music starts in the second half. Typical cases where something like this can be found is in piano music where a voice continues on the other staff:

Okay, now I understand the Sibelius data, thanks.

> Inputters might be "disturbed" by the grayed out hidden rests and delete them altogether. This is not good practice, but it can be found in the wild and we should make sure that it does not crash Sibmei.

Of course, and it works currently in this case. But there is a difference between prevent crashes and produce good data. But in most cases, it is just junk in junk out... We cannot prevent this and shouldn't do this.
 
> If it helps to make the intention of the test clearer, I can turn the test file into a grand staff and use the situation illustrated above instead of the strange bar.

I don't think it's necessary to change the test. You already mention the oddness in the file. I was just wondering why this is degenerate... ;-)

> > But one question before we merge: Do we need `@staff` and `@tstamp` for clefs?
> 
> Oh, there are a lot of attributes that we don't really need, I think. (Do we need `@tstamp.real` for example? Not sure for what reason that was added.) Any specific reason to remove `@staff` and `@tstamp`?

`@tstamp.real` used to be `@tstamp.ges` before I updated to MEI 4.0. I already wondered why there is so much output regarding midi-related information. I always throw them out... But I assume it is exported because at least one person is using this and I am a bit afraid of throwing this away in general.
But having `@staff` and `@tstamp` is a little bit different. This information has nothing to do with any midi stuff and especially `@staff` seems a little bit redundant to me.
Anyway, I have no problem with still having those attributes if there might be a possible use of them. I just wanted to ask...

If you don't want to throw them out, I have no problem with merging.

ahankinson:
The MIDI output is there because sometimes it's the only way to represent certain types of information coming out of Sibelius, and I wanted the output to be as comprehensive as possible. If you don't need it it's safe to ignore it, but it can be used as hints for playback devices (for example).

th-we:
Hi @ahankinson, good to see you're still watching this "baby" of yours despite your MEI sabbatical!

As it's always easier for users to remove data than to add it back, I'd say just keep all the attributes as they're not wrong. 

To prevent future irritations, I'll update the pull request with a modified test case anyway.

th-we:
@annplaksin Updated the test file. Can you give it a go?

PR #77, opened 2017-03-25
Support @plist, @endid etc. on spanners

th-we:
For me this is especially important for `<octave>` to carry over explicitly what notes are affected by the `<octave>` transposition, but should also be useful for any other spanner/line-like object.

Previously, no info where a line ends was encoded.

[ottavas.sib.zip](https://github.com/music-encoding/sibmei/files/869919/ottavas.sib.zip)

### DIFF
--- a/src/ExportProcessors.mss
+++ b/src/ExportProcessors.mss
@@ -822,4 +822,76 @@ function ProcessSymbol (sobj) {
         }
 
     }
+} //$end
+
+function ProcessActiveSpanners (bobj, meiId) {
+    //$module(ExportProcessors.mss)
+    /*
+        This function does two things:
+         * Adds the bobj's ID to any active spanner's plist.
+         * When we've gone beyond the end of an active spanner, adds 
+           start/end/plist/duration info.
+        This function is also called with `nil` arguments when we've 
+        reached the end of the score.
+    */
+  
+    if (bobj != null)
+    {
+        barNum = bobj.ParentBar.BarNumber;
+        pos = bobj.Position;
+        bobjVoiceNum = bobj.VoiceNumber;
+        minStaffNum = bobj.ParentBar.ParentStaff.StaffNum;
+        maxStaffNum = minStaffNum;
+    }
+    else
+    {
+        minStaffNum = 1;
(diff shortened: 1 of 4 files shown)


th-we:
Spotted some remaining problems I'll have to fix.

src/ExportProcessors.mss
```diff
@@ -822,4 +822,76 @@ function ProcessSymbol (sobj) {
...
+} //$end
+
+function ProcessActiveSpanners (bobj, meiId) {
+    //$module(ExportProcessors.mss)
+    /*
+        This function does two things:
+         * Adds the bobj's ID to any active spanner's plist.
+         * When we've gone beyond the end of an active spanner, adds 
+           start/end/plist/duration info.
+        This function is also called with `nil` arguments when we've
```
ahankinson: Why?

th-we: The original motivation for this was that it's not necessarily straightforward to find out whether a note is affected by an `<octave>` element. An explicit `@plist` makes this crystal clear.

`@endid`/`@tstamp2` is very sensible for many types of line like elements. `@endid` can only be added after we've processed the element a line ends on. That's why this commit keeps track of "active spanners".

ahankinson:
Do you have an update on this, @th-we ? I was going to merge it if you've addressed the problems you said were remaining.

th-we:
@ahankinson The issues I had back then turned out to be unrelated to this commit. (The reason why I fell silent after this is that we couldn't use this code as planned and I had to invest all my energy in a plan B.)

If there are no objections to putting all these attributes on spanners, then this could be merged. I found them very helpful for further processing the output MEI.

rettinghaus:
@th-we What about this PR?

annplaksin:
With the MEI 4.0 changes I added `@startid` and `@endid` to most of the lines, including trills and octava lines, if the line object is assigned to a specific layer.
I don't know about any current support of `@plist`. Would that be neccessary for anyone?

rettinghaus:
I guess `@plist` isn't used often. So we could postpone this. 
@th-we If you don't object, i'll close this PR.

th-we:
O.K., I'll just close it myself.

PR #78, opened 2017-03-25
Treat line-like and symbol-like trills as separate things

th-we:
This builds on top of #77 so that trills also get `@tstamp2`, `@plist`, `@endid` etc.  Simpler than having a common function that tries to handle both types of trills.

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -1430,9 +1430,8 @@ function GenerateLine (bobj) {
         }
         case ('Trill')
         {
-            line = GenerateTrill(bobj);
-            // NB: Return here since the trill already has its properties set.
-            return line;
+            line = libmei.Trill();
+            libmei.AddAttribute(line, 'extender', 'true');
         }
         case ('Line')
         {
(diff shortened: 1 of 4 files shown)


th-we:
#77 still needs some fixes, until then this will have to wait.

ahankinson:
Why treat line and symbol trills differently? Won't this just cause two different encodings of the same thing? (Something I would like to avoid, if possible).

th-we:
@ahankinson What do you mean by "two different encodings"?  I assume you don't mean that two `<trill>` elements might be created where there was only one Sibelius object, right?  The old `GenerateTrill()` function did nothing that wouldn't be covered by the generic handling of symbols and spanners.

What I think would be more sensible is to generalize repetitive code in `ProcessSymbols()`, but that's maybe a different issue.

src/ExportGenerators.mss
```diff
@@ -1430,9 +1430,7 @@ function GenerateLine (bobj) {
         }
         case ('Trill')
         {
-            line = GenerateTrill(bobj);
-            // NB: Return here since the trill already has its properties set.
-            return line;
+            line = libmei.Trill(bobj);
```
ahankinson: You don't pass in the `bobj` to an element constructor...

th-we: That's of course correct.

ahankinson:
I'm just not sure of the reason for this pull request. It seems to me that this would mean that consistent encoding of trill elements should go through the `GenerateTrill` method, since otherwise it would be more likely that handling trills would become inconsistent, meaning one type of trill would have functionality that is not reflected in the second type.

It just seems like this pull request is a step backwards in terms of functionality and consistency. (It's also curiously missing the functionality for attaching a `startid` to trills?)

th-we:
Before this pull request, trill lines would not get `@endid`/`@tstamp2` or `@plist`. `ProcessActiveSpanners()` adds this info.

`@startid` is added by `AddBarObjectInfoToElement()` anyway, so we don't need trill specific code here.

I agree that if we had a chance to add `@accidupper` or `@accidlower`, then we should have common function that handles this. But this could be called from the two different places.  At the moment I don't think there's a standard way of representing this information in Sibelius, though.

rettinghaus:
We decided that this needs more work. So closing this for now.

PR #79, opened 2017-04-05
Trust Sibelius' IsAccidentalVisible

th-we:
Sibmei's algorithm for determining accidental visibilty has its imperfections, especially in situations with ties over barlines. `IsAccidentalVisible` reports this correctly, though, so why not always use, not just when it's true?

[accidental-visibility-on-ties-over-barlines.sib.zip](https://github.com/music-encoding/sibmei/files/895891/accidental-visibility-on-ties-over-barlines.sib.zip)


ahankinson:
Largely because accidental visibility was the hardest part of writing this plugin, and I only found out about this particular feature by accident and couldn't believe that it really existed, so I had to make it over-complicated. :)

PR #83, opened 2017-06-07
Implement figured bass

th-we:
This feature was created to fit the needs of a project at the [ÖAW](https://www.oeaw.ac.at/ikm/home/) (Austrian Academy of Sciences). Some functionality, e.g. "progression lines", is still very basic.

Example file: [figbass_symbols.sib.zip](https://github.com/music-encoding/sibmei/files/1059303/figbass_symbols.sib.zip)

The approach we took is to normalize all figures to digits and Unicode accidentals, even for strike-through symbols. This has two reasons:

* Using SMuFL characters from the private use area directly in the text might lead to strange results.
* We hope the normalized data is more accessible to analysis, database applications and the like.

To not lose the information about special figures, we created `<symbolDefs>` for each encountered SMuFL character and pointed to it using `@altsym`.

### DIFF
--- a/src/ExportConverters.mss
+++ b/src/ExportConverters.mss
@@ -985,6 +985,7 @@ function ConvertFbFigures (fb, bobj) {
 
     n = 1;
     currentLine = '';
+    altsym = null;
     components = bobj.TextWithFormatting;
 
     // Strangely, components.Length is null, so we can't use a for loop.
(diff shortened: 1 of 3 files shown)


ahankinson:
Thanks!

th-we:
@ahankinson The ÖAW wants to ask for being mentioned as the sponsor of this feature in the next release notes. Would you be so kind as to take care of that?

ahankinson:
Yes, of course.

th-we:
I don't mind! In any case, maybe not right now as there *might* be some minor additions coming for this feature.

ahankinson:
OK, I will not do it now then.

th-we:
We will not add anything to this feature at this point, so if you think there should be a release, I have no objections. If you do so, please don't forget to mention the ÖAW as sponsor of this feature (Notengrafik Berlin is only the contractor in this case).

Issue #90, opened 2018-06-13
Performed duration of a note within <tuplet> has a larger value than outside <tuplet>


martha-thomae:
The performed duration (`@dur.ges`) of a note has a larger value when it is part of a `<tuplet>` than when it stands alone, which I think is opposite of the expected behaviour. Here I give an example.

[performed-duration.zip](https://github.com/music-encoding/sibmei/files/2099744/performed-duration.zip)

As you can see in this example, the quarter notes have a `@dur.ges="256p"` and the eighth notes have a `@dur.ges="128p"` when they are not part of a `<tuplet>`. However, when these notes are children of the `<tuplet>` element they performed duration increases to `@dur.ges="384p"` for the quarter note and `@dur.ges="129p"` for the eighth note. I believe the correct values would be `@dur.ges="171p"` for the quarter note and `@dur.ges="85p"` for the eighth note since now there are three notes in the place of two (given that `<tuplet @num="3" @numbase="2">`).

I think this is due to the **line 953** in `src/ExportGenerators.mss`:
`floatgesdur = (pnum & '.0' / pden & '.0') * dur;`
which I believe should be the other way around:
`floatgesdur = (pden & '.0' / pnum & '.0') * dur;`

What are your thoughts?
I can make the change and the pull request if you agree.

ahankinson:
That sounds reasonable. Good job catching that! Please file a PR.

PR #91, opened 2018-06-15
Fixed: Gestural duration of notes within a tuplet.

martha-thomae:
Previously the gestural duration of a note had a larger value when it was part of a `<tuplet>` than when it stood alone. This commit fixes this by inverting the ratio by which the duration is multiplied. The gestural duration (`@dur.ges`) of the stand alone note is now multiplied by the ratio `@numbase/@num` to get the `@dur.ges` of the `<note>` element when it is part of a `<tuplet>`.

Fixes https://github.com/music-encoding/sibmei/issues/90

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -962,7 +962,7 @@ function GenerateNote (nobj) {
         ptuplet = nobj.ParentNoteRest.ParentTupletIfAny;
         pnum = ptuplet.Left;
         pden = ptuplet.Right;
-        floatgesdur = (pnum & '.0' / pden & '.0') * dur;
+        floatgesdur = (pden & '.0' / pnum & '.0') * dur;
         gesdur = Round(floatgesdur);
     }
     else


ahankinson:
Could you check that it works correctly with the MEI test set?

http://ddmal.github.io/mei-test-set/MEI-test-set/tuplets.html

https://github.com/DDMAL/mei-test-set/tree/gh-pages/MEI-test-set/sibelius-files/tuplets

Thanks!

martha-thomae:
Hi Andrew,
By checking that they work correctly with the MEI test set, do you mean exporting these Sibelius files into CMN MEI and then render them in Verovio to see they are correct by comparing them to the version given in http://ddmal.github.io/mei-test-set/MEI-test-set/tuplets.html? Or is there a mechanism other than visual inspection?

ahankinson:
I'm afraid visual inspection is the only way

martha-thomae:
Ok, I did it. I went through each of the [tuplet files](https://github.com/DDMAL/mei-test-set/tree/gh-pages/MEI-test-set/sibelius-files/tuplets), used the edited SibMEI plugin to get the CMN MEI files, and rendered these in the [Verovio MEI Viewer](https://www.verovio.org/mei-viewer.xhtml). There are a few of these files that, when rendered in Verovio, don't look as in the [github pages](http://ddmal.github.io/mei-test-set/MEI-test-set/tuplets.html). But if I take all these MEI files from the [github pages](http://ddmal.github.io/mei-test-set/MEI-test-set/tuplets.html) (generated with the last SibMEI release of 2.1.0) and render them in the same Verovio MEI Viewer (which is version 2.0.0-dev-7b6caf4) they have the same "mistakes" as the ones newly generated with the SibMEI plugin I edited.

I also checked the original MEI files (the ones generated with SibMEI 2.1.0) against the ones that I just generated (with the edits I did to SibMEI) and the only difference was the `@dur.ges` of the notes within the tuplets. I used the FileMerge app to compare them. So, in theory, the result of the SibMEI plugin is the same before and after my small edit, the only thing that changed is the performed duration, which is what we expected. So, I would say that it does work correctly with the MEI test set, but I would like to hear what you have to say about it.

martha-thomae:
FYI, the files that did not looked the same were the ones involving:
- Bracket placement in nested tuplets (`bracket-placement.mei` and `nested-tuplets.mei`)
- The ones in which a ratio was supposed to be represented by a single number (`ratios.mei` and `ratios-no-literal.mei`)
- And the ones with articulations. Specifically, the slurs within a tuplet were not represented. Ties and staccatos are actually shown. (These files are `tuplets-articulation.mei` and `tuplets-articulation-2-parts.mei`)

As I said before, these are the items in the test set that look different in github pages when compared to their rendition using the Verovio MEI Viewer (regardless if the files used for Verovio came from the SibMEI 2.1.0 release or the edited SibMEI).

rettinghaus:
@martha-thomae you should try it with Verovio 1.1.6, because Verovio 2 is based on MEI 4 which uses a changed approach on gestural durations. 

http://music-encoding.org/guidelines/dev/data-types/data.duration.gestural.html

martha-thomae:
Thanks @rettinghaus! I will try that now.

martha-thomae:
Hi @rettinghaus and @ahankinson! I already tried these files with the Verovio 1.1.16 release, and I am facing the same issues. Even the original files are rendered incorrectly. I attach here one example, this is `bracket-placement.mei` of the [test set](http://ddmal.github.io/mei-test-set/MEI-test-set/tuplets.html) but rendered in Verovio 1.1.16. As you can see the rendition shown in the previous link and the one attached here do not look the same (and these are the original files, generated with the original SibMEI 2.1.0).

screen shot 2018-06-19 at 1 56 58 pm

martha-thomae:
@rettinghaus I agree.
The `alignment.mei` works perfectly, both the SibMEI 2.1.0 and the SibMEI I edited generate the same result and they both look as they should in Verovio.

**@ahankinson could we accept this pull request?** All the output MEI files from the SibMEI edited here are the same as the original MEI files coming from SibMEI 2.1.0.

ahankinson:
I will accept it, but if you have a moment could you please update the MEI in the MEI test set? I use the 'batch' plugin that comes with SibMEI to convert an entire folder. Thanks!

ahankinson:
Also, for future reference, it's better to create your own dedicated 'fix' branch for specific issues, and to do your pull requests from this. You put in this pull request from your 'master' branch on your fork, but it would be better to make a dedicated branch ('fixed-tuplet-duration') and then to issue your pull request from that branch to the master branch on SibMEI.

ahankinson:
Oh, and thank you for the fix! That's really helpful.

martha-thomae:
@ahankinson yes, I will update the MEI test set as soon as I have the time.

> Also, for future reference, it's better to create your own dedicated 'fix' branch for specific issues, and to do your pull requests from this. You put in this pull request from your 'master' branch on your fork, but it would be better to make a dedicated branch ('fixed-tuplet-duration') and then to issue your pull request from that branch to the master branch on SibMEI.

Noted, I will do that next time. Thank you!

And thanks for accepting the pull request!

PR #101, opened 2019-03-25
Escaped characters in elision

th-we:
`libmei.GetText(sylel)` returns an already entity-escaped string. We want to work on the "raw" string instead or we'd double-escape characters.


src/ExportProcessors.mss
```diff
@@ -292,12 +292,10 @@ function ProcessLyric (lyricobj, objectPositions) {
             }
         }
 
-        syltext = libmei.GetText(sylel);
```
ahankinson: You probably don't need the statement in line 254 above now...

th-we: You mean the `libmei.SetText()`? We still need it. Only if we have an underscore in the syllable we re-set the text. We could move it to an `if`/`else` if you think that makes clearer what's happening:

```       
if (utils.Pos('_', syl.Text) <= -1) {
    libmei.SetText(sylel, syl.Text);
}
else
{
    // Syllable elision. split this syllable element by underscore.
    syllables = MSplitString(syl.Text, '_');
    sylarray = CreateSparseArray();

    // reset the text of the first syllable element to the first half of the syllable.
    libmei.SetText(sylel, syllables[0]);
    ...
```

ahankinson: Nah, we can avoid that. I'll merge it.

PR #112, opened 2019-07-22
Fixed gulpfile

annplaksin:
After the last changes for gulp version 4.0.2 the watcher triggered only once.
To fix this, I adjusted the callback in the build function.
Also, I switched from gulp task() to exported functions since this is now the recommended method; see [task()](https://gulpjs.com/docs/en/api/task).

I am not 100% sure if the callback fix is the best way to do it; it works now...
@th-we Could you please have a look? -> [gulp Async Completion](https://gulpjs.com/docs/en/getting-started/async-completion) 
It seems like you could use promises without the q package, but I am familiar enough with that.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -8,7 +8,7 @@ var l = require('fancy-log');
 var Q = require('q');
 var plgconf = require('./plgconfig');
 
-gulp.task('develop:build', function(callback)
+function build()
 {
     var deferred = Q.defer();
     l.info(c.blue('Copying "linked" libraries'));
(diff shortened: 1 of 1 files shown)


th-we:
I am not an experienced Gulp user.  I checked out your branch and found that your code still only triggers once.

annplaksin:
Really? O no, this sounds totally stupid... I tried to get into gulp a bit more, but it is still confusing.
Is you gulp version 4.0.2 as well?

th-we:
Ehm, not sure...

```
$ gulp --version
CLI version: 2.2.0
Local version: 4.0.2
```

annplaksin:
Hmm... jep, the version is 4.0.2
And okay, this is very strange right now... it doesn't work anymore for me either...
Then I really should try this again... sorry for the mess!

annplaksin:
@th-we Try again please!

th-we:
Works for me!

annplaksin:
Jay! 👍

PR #113, opened 2019-07-23
Fixed test: creates correct character entities on elisions

annplaksin:
`syls[13].firstChild._data` returned undefined, used `syls[13].firstChild.F` instead.
**F** is really not a very intuitive variable, but it seems like this is the way to get the text of an element according to the JSON returned when querying for the whole element...


th-we:
Kurios.  I have no idea why I used `_data` – it should be the standard DOM `textContent`.  Can you change that?

annplaksin:
Changed... but where do I find some documentation for that?

th-we:
https://www.w3.org/TR/2004/REC-DOM-Level-3-Core-20040407/core.html#Node3-textContent

PR #108, opened 2019-07-01
SymbolItems with @startid and fixed @startid/@endid for slurs within the same measure

annplaksin:
This PR provides the export of ornaments with @startid: closes #62 

Additionally the structure of GenerateLayer() is redone to fix the export of @startid and @endid for slurs.
It must be ensured that Line objects are processed after NoteRest obejcts to be able to find the corresponding notes. Usually, if more than one object is at the same position, the NoteRest object is the last one.

PR #114, opened 2019-08-26
Anna's @startid/@endid pull request rebased/amended supporting cross-measure slurs

th-we:
I fixed two minor things in two commits from @annplaksin's PR #108 and added support for slurs across measures. (I.e. this supersedes #108.)

Anna, could you check this?


src/Utilities.mss
```diff
@@ -78,6 +78,72 @@ function SimpleNoteHash (nobj) {
...
+    {
+
+        barNumber = barInfo.BarNumber;
+        staff = barInfo.ParentStaff;
+    }
+    else
+    {
+        barNumber = barInfo;
+    }
+    if (null = staff) {
```
annplaksin: This might be petty,  but it might aid the readability to not switch the agruments from the statement above...

annplaksin:
I just checked your branch out for testing after a short look on the code, but I ran into the the already fixed sib-test crashes. 
When I look at the TestExportGenerators file in your branch at Github, there is still the old bug in line 55. It might be totally possible, that only I am running into that problem again, or it is my fault, that it is still occuring.
Could you please make sure, that this branch is based on a version with working tests? 
(Then I could try again and see how confused I am... )

src/Utilities.mss
th-we: I adopted the habit of putting null on the left side in ManuScript because `null=CreateArray()` is false as expected, but `CreateArray()=null` is true for some reason (a general problem with TreeNodes). But comparison with null is something very hairy anyway (`0=null`, `''=null`, `' '=null`), so I'll change that.

annplaksin: I was just not aware of the habit, sorry...
If you have good reasons and it works, there is no need to change that.

th-we:
@annplaksin I only verified the new test works and commented out everything else.  I'll work on that!

annplaksin:
I tested it again and it worked this time after I merged your latest change into my local branch....   
Only my fix of the elision test is still lacking. If this is changed, the PR should be ready to merge.

annplaksin:
Err... sorry for my constantly growing confusion... I really lost my head while writing my last comment.
Of course the fix of the elision test is not a problem... 
Sorry!

th-we:
I realize I wrote incomprehensible comments for the `LayerHash()` function. I tried to make them slightly less incomprehensible now.  Does the approach of calculating this "hash" make sense to you or do you think this is getting too complicated? My goal was to make writing to and reading from `Self._property:SlurResolver` easy.

annplaksin:
I'll ask @Prohl about that... ;-)

annplaksin:
It took us quite some time yesterday to think through it. 
Since I understand it correctly, that you only want to identify the layers with that `LayerHash()` function, it should work, I assume...
I didn't fully get why you use the bar number and staff.BarCount as well, since in most cases the BarCount of staffs would not differ, but in my opinion, this is nothing that prevent us from merging...

rettinghaus:
Is everyone happy with this?

annplaksin:
I am happy with this

PR #120, opened 2019-11-06
MEI 4.0.1 valid output

annplaksin:
This PR provides the necessary changes for a MEI 4.0.1 valid sibmei output.

* Updated libmei
* Header has changed
* Changed note attributes
* Changed mordent and turn attributes
* Changed notehead shapes
* Added `@startid` and `@endid` to most lines whenever possible
* Changed `@accid.ges` values
* Code clean up
* Added tests for most features

@th-we and @ahankinson , would you please have a look?


ahankinson:
Yes, if we want one last MEI 3.0 plugin release then `develop` should be merged to `master` before this gets merged to `develop`.

I would then make a new branch from master (`legacy-2x`? Open to suggestions...) and maintain that separately so that if we need to make changes to the MEI 3.0 support we can.

I would also suggest that we skip a version number, and that when @annplaksin 's contributions are merged we make it version 4.0, so that from now on the sibmei version tracks the MEI version.

rettinghaus:
@th-we do you have something left to merge or could you make a release?

th-we:
Superb, @annplaksin! I'll read through and add some comments.

src/ExportConverters.mss
```diff
@@ -641,7 +610,7 @@ function ConvertNoteStyle (style) {
     {
         case (CrossNoteStyle)
         {
-            noteStyle = 'cross';
+            noteStyle = 'x';
```
th-we: I think we should only assign `'x'` if `noteobj.ParentNoteRest.Duration <=256`, otherwise `'diamond'`

src/ExportConverters.mss
```diff
@@ -653,18 +622,21 @@ function ConvertNoteStyle (style) {
             // and does not differentiate in the
             // head style, so we have to choose one
             // or the other.
-            noteStyle = 'cross';
+            noteStyle = 'x';
         }
         case (BlackAndWhiteDiamondNoteStyle)
         {
+            // There is not filled diamond in data.HEADSHAPE.list
```
th-we: I think we should get rid of the non-standard names we currently use. In this case, we should use `'diamond'` if `noteobj.ParentNoteRest.Duration > 256` and otherwise [`'#xE0DB'`](https://smufl-browser.edirom.de/index?glyphname=noteheadDiamondBlack). In `GenerateNote()`, we'd then have to check for leading `'#x'` and set `@head.auth="smufl"` accordingly.

th-we:
I think this is really good. Some things I found that could be improved were not really related to the changes made in this pull request. (Except maybe the [>150 lines case](https://github.com/maxweberstiftung/sibmei/commit/327ad4fb05917e2b46b85a9609da2423c6121463#diff-d9e47d87f7a709a9aaefe8d86469f103R1526), but as long as it does its job...) So I'm ready to merge and release.

Further steps would be:
* Create a legacy branch "legacy-mei3", if you agree with me that the MEI version is more important than the Sibmei version.
* Merge this PR into develop
* Replace all sibmei2 occurrences with sibmei4

rettinghaus:
@th-we The latest improvements prior to this PR should go into version 2.3.0, then we're ready to go for MEI4.

th-we:
@rettinghaus Of course – I merged all changes from develop into master and drafted [release notes](https://github.com/music-encoding/sibmei/releases/tag/untagged-635bbb0a35ec10acf47c). Could you review them?

[BTW: @annplaksin, can you see the release notes? If not, @ahankinson, would you give Anna write access? As we (Anna and me) are the most active contributors at the moment, it would be nice if we could review and merge each other's pull requests.]

rettinghaus:
@th-we Looks good to me, please go ahead.

th-we:
@ahankinson Any objections to release 2.3.0?

annplaksin:
@th-we Nope, I get a 404 when I click the link.

Thanks a lot for your feedback!

So, just for mentioning. I think as well, there might be some approvements in handling the shapes of note heads possible. Because I tried to take not so much more time for this PR, I only made some very obvious changes. And at some point, because the head shapes in Sibelius are depending a lot on the duration, I thought it was not so bad to use non-standard terms in that state of export.

ahankinson:
What's the status on this? Can it be merged?

ahankinson:
Great. I think the sequence is:

 - Release 2.3.0 (currently drafted)
 - Merge this to develop
 - Merge #122 to develop?

Or is there some other order? @rettinghaus ?

th-we:
I'll release and then merge.

PR #121, opened 2019-11-10
Update Sibmei version for release

th-we:
@ahankinson Can we just remove the version field in package-lock.json or do we have to keep in in sync with the one in package.json?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "sibmei2",
-  "version": "2.2.0",
+  "version": "2.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sibmei2",
-  "version": "2.2.0",
+  "version": "2.3.0",
   "description": "",
   "main": "plgconfig.js",
   "directories": {


ahankinson:
I think we need to keep it in sync? Not sure

th-we:
O.K., I updated both files now. Shall I go ahead and release?

ahankinson:
Should this not be merged to develop, and then develop to master?

ahankinson:
If it's a matter of permissions, I've just given you write permission

th-we:
This is not a matter of write permissions – just flawed thinking on my side.  (Forgot to change the version numbers before merging into master.)  Of course you're right – this should go into develop first and then into master.  Good I made this a pull request instead of pushing to master right away.

rettinghaus:
@th-we I just noticed, that you haven't changed `GLOBALS.mss`, so in the release there is still version 2.2.0 indicated in the MEI header.

th-we:
Ehm, right. I think I need a checklist to work through before releases. Options now:

1. Update version number to 2.3.0 in a new commit
  * Make the tag v.2.3.0 point to that commit
  * Update ZIP file on the release page
2. Create a version 2.3.1 that has proper version numbers

I'd go for option 1. Opinions?

rettinghaus:
Option 1 sounds good.

th-we:
On it goes here #123
Sorry for the trouble. Already updated the zip file with the proper version number.

PR #123, opened 2019-11-20
Update version to 2.3.0 in GLOBALS.mss

th-we:
O.K., now this needs to be merged to master and develop, right?


th-we:
O.K., I also updated the tag, it seems to have worked.

PR #126, opened 2019-12-09
Fix elisions

annplaksin:
If there is currently an elision binding together two words with more than one syllable, the `@con='b'` will be overwritten by the middle of word condition.
E.g.: "Ky - ri - e_lei - son"

This PR currently provides an updated test and prevents the replacement of the connector but it is far from mergable right now.
It still lacks a connector of the second syllable and proper `@wordpos` attributes. 

My question, before I proceed working on that problem:
* Could we assume that an elision always the placing of more than one syllables on a note? (In linguistics an elision is the deletion of a sound within a word or phrase)
* Could we then assume, a syllable with `@con='b'` is a word boundary?

If both questions could be answered with **yes**, it would be neccessary to rework the for-loop and split the detection of elisions and adding `@wordpos` and `@con` attributes.
I will work on that, but before messing around, I would like to ask you first...


annplaksin:
Given the assumption, that an elision is always the end of a word, the work is done:
* An elision connecting two word with more than one syllable gets `@wordpos` and `@con` attributes
* Added a case with an elision at the end of a word to the test

@th-we could you please have a look? 
If there are no objections, I would consider to include it in version 4.0.0

rettinghaus:
Closing, as proposed by @th-we

PR #127, opened 2019-12-09
More info about tests in README

th-we:
Is more detailed documentation of the tests needed?


annplaksin:
I added the link to the mocha docu. That should be fine for now...
Maybe we find some time in the future to discuss neccessary topics in a sibmei documentation a bit more in detail.

PR #130, opened 2020-03-12
bring develop ahead of master

rettinghaus:
Last PR by @th-we went directly to master, if we do not revert that one we need to merge changes back to develop.


annplaksin:
Yes, good that you noticed that. Having those PRs in develop seems necessary to me.

th-we:
Oh, what a mess I made again.  What about this:  I reset master 45f38c0 by force push, and we put everything back on develop?

PR #137, opened 2020-04-14
improve tuplet attributes

rettinghaus:
Minor improvement on tuplet attributes. 
NB. Why do we have an unused `ConvertTupletStyle` function?

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -1433,7 +1433,12 @@ function GenerateStaffGroups (score, barnum) {
 function GenerateTuplet(tupletObj) {
     //$module(ExportGenerators.mss)
     tuplet = libmei.Tuplet();
+    dur = tupletObj.PlayedDuration;
+    meidur = ConvertDuration(dur);
 
+    // we simply assume no dotted duration for now
+    libmei.AddAttribute(tuplet, 'dur', meidur[0]);
+    libmei.AddAttribute(tuplet, 'dur.ppt', dur);
     libmei.AddAttribute(tuplet, 'num', tupletObj.Left);
     libmei.AddAttribute(tuplet, 'numbase', tupletObj.Right);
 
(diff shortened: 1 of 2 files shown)


src/ExportGenerators.mss
```diff
@@ -1422,7 +1422,12 @@ function GenerateStaffGroups (score, barnum) {
 function GenerateTuplet(tupletObj) {
     //$module(ExportGenerators.mss)
     tuplet = libmei.Tuplet();
+    dur = tupletObj.PlayedDuration;
+    meidur = ConvertDuration(dur);
```
th-we, line 1426: `PlayedDuration` should not be used here because it can have funny values like `341` that don't make sense for `ConvertDuration()`.

I'd rather not add `@dur` unless we do it properly (I'm also ). Any argument for having it anyway?

src/ExportGenerators.mss
```diff
@@ -1422,7 +1422,12 @@ function GenerateStaffGroups (score, barnum) {
...
 
+    // we simply assume no dotted duration for now
+    libmei.AddAttribute(tuplet, 'dur', meidur[0]);
+    libmei.AddAttribute(tuplet, 'dur.ppt', dur);
```
th-we, line 1430: Should that be `@dur.ppq`?

src/ExportGenerators.mss
rettinghaus: That's odd. But maybe I simply missunderstood the ManuScript Manual here.

src/ExportGenerators.mss
rettinghaus: absolutely right!

annplaksin:
@rettinghaus Could you please add some test files here as well?

rettinghaus:
@th-we this needs to get in, to fix outdated attributes

rettinghaus:
superseded with #199

PR #139, opened 2020-04-25
add text test file

rettinghaus:
This adds a test file for different text styles and offsets. 
I noticed, that Sibelius places everything above the staff and moves it below with offsets. The exported MEI files don't carry the `@place` attribute. Should we add `@place="above"` to every control element, or should we add a function that checks the vertical offset, adds `@place="below"` eventually and substract the staff height from the offset value?


th-we:
Tricky question. You're absolutely right, the `@vo` we're currently writing does not make sense.  Sibelius' `Dy` and MEI's `@vo` are actually built on opposing concepts:

*  [`@vo`](https://music-encoding.org/guidelines/v4/attribute-classes/att.visualoffset.vo.html#attributes) is a relative offset, an "adjustment of a feature's programmatically-determined location", i.e. the position the program would put the element at is taken as the starting point which is then tweaked by the value of `@vo`.
* `Dy` defines an absolute position (with 0 being the center of the staff). If magnetic layout is turned on, then the program takes that position as a starting point and tweaks it.

`@vo` has the inherent flaw that the "programatically-determined location" differs depending on the application. `Dy` has the problem that we have no way of knowing what offset magnetic layout adds.

Conclusion: Don't write `@vo` at all?? It would be good if we could write `@place`, but even that's impossible to determine reliably without reverse-engineering magnetic layout. Check the following example:

grafik

All the texts have `Dy=-32`, but depending on the context, they're either placed above or below by magnetic layout. While `Dy=-32` is a position within the staff (*g'* line in treble clef), even if `Dy` is 96 (*a''* ledger line, i.e. clearly above the staff), the text could be pushed below the staff by the context:

grafik

rettinghaus:
@th-we Yeah, pretty tricky. In tendency we could take as rule of thumb negative Dy values as `place="below"` and positive ones as `place="above"` and then drop `@vo` from export. It is clear form your examples that using MEI's `@vo` for Sibelius' `Dy` values directly doesn't make much sense.

th-we:
What we can do:

* If magnetic layout is off, then we can definitely tell if `@place` should be `below` or `within`. To decide if the text is truly above, we'd have to know if it has wrapped lines, which we can't really tell because we can't find out if the text is wrapping automatically, like this:
    grafik
    But I'd be fine with just setting `@place="above"` if the text starts above the system.
* With magnetic layout turned on, we can experiment for which values of `Dy` it's safe to assume the text is above or below. For any other values, we just don't write the `position` info. Better to omit than to write potentially false information.

I'm wondering if `@ho`/`@vo` should be deprecated altogether in MEI. It's so highly application dependent that if you really need to influence the rendering of a specific application, it would be better to create a customization that actually represents the features of that application properly.  Say `@lilyOnceOverride` and `@lilyTweak` for MEILER, or `@sibDy` for a (theoretical) Sibelius import.

rettinghaus:
During the developer workshop we decided to combine these test cases with the test file already in place. 

Closing this PR.

PR #140, opened 2020-04-28
set vu.height and spacing attributes

rettinghaus:


th-we:
This raises a general question: Why do we have `@vu.height` on `<scoreDef>`, but not on `<staffDef>` in MEI?  Different staff sizes can be used for different staffs, and staff size can even be changed between systems. Unfortunately, we don't have access to that information with ManuScript.

rettinghaus:
Simple answer `@vu.height` gives the general "virtual unit" in MEI that is used for the whole document. Staff size usually is eight times vu. However, you may set different staff sizes with `staffDef@scale`.

rettinghaus:
… and 'StaffSize' gives Sibelius' *normal* staff size, isn't it?

PR #141, opened 2020-04-29
fix rests

rettinghaus:
This PR
* removes the unavailable `visible` attribute on `rest`

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -809,7 +809,7 @@ function GenerateNoteRest (bobj, layer) {
 
     if (bobj.CueSize = true and libmei.GetName(nr) != 'space')
     {
-        libmei.AddAttribute(nr, 'fontsize', 'small');
+        libmei.AddAttribute(nr, 'cue', 'true');
     }
 
     if (bobj.Hidden = true and libmei.GetName(nr) != 'space')
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -970,12 +970,7 @@ function GenerateRest (bobj) {
 
     if (bobj.CueSize = true and name != 'space')
     {
-        libmei.AddAttribute(r, 'fontsize', 'small');
-    }
-
-    if (bobj.Hidden = true and name != 'space')
-    {
-        libmei.AddAttribute(r, 'visible', 'false');
+        libmei.AddAttribute(r, 'cue', 'true');
     }
 
     if (bobj.Color != 0 and name != 'space')
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -970,7 +970,7 @@ function GenerateRest (bobj) {
 
     if (bobj.CueSize = true and name != 'space')
     {
-        libmei.AddAttribute(nr, 'fontsize', 'small');
+        libmei.AddAttribute(r, 'fontsize', 'small');
     }
 
     if (bobj.Color != 0 and name != 'space')
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -809,7 +809,7 @@ function GenerateNoteRest (bobj, layer) {
 
     if (bobj.CueSize = true and libmei.GetName(nr) != 'space')
     {
-        libmei.AddAttribute(nr, 'cue', 'true');
+        libmei.AddAttribute(nr, 'fontsize', 'small');
     }
 
     if (bobj.Hidden = true and libmei.GetName(nr) != 'space')
@@ -970,7 +970,7 @@ function GenerateRest (bobj) {
 
     if (bobj.CueSize = true and name != 'space')
     {
-        libmei.AddAttribute(r, 'cue', 'true');
+        libmei.AddAttribute(nr, 'fontsize', 'small');
     }
 
     if (bobj.Color != 0 and name != 'space')


rettinghaus:
Test file: [rests.zip](https://github.com/music-encoding/sibmei/files/4553103/rests.zip)

annplaksin:
@rettinghaus `@cue` should be used for small symbols? Maybe I didn't get any hint by reading the documentation for using this:
> Attributes that describe "cue-ness".

I'm sorry that I didn't know any better. It would have been easier if someone would have informed me.

This change would as well make it neccessary to rewrite two tests of my MEI 4.0 update, since I tested for `@fontsize`. As well, the encoding would need to be adjusted for notes as well.
See commit https://github.com/music-encoding/sibmei/commit/ee9c69d4b30ebc57c172ff8e26a5c25ca9a621cf

Adjusting the existent test shouldn't be a huge task. Just the attribute that is tested for, must be changed. 
@th-we  & @rettinghaus Should we adjust the test or make a new one for the rests in general, that includes a test for `@loc`. For the latter, I would appreciate a description of the expected output since it doesn't make sense to write a test that fits the actual output.

rettinghaus:
I just guessed, that `CueSize` implies, that this note is a cue note. 
We don't have `@fontsize="cue"`, and what would you do with `GraceNoteSize` and `CueGraceNoteSize`? If we'd prefer to you fontsize, it should take the values from `CueNoteScale` and `GraceNoteScale` as `data.PERCENT`, instead of `data.FONTSIZETERM`. 
What's your take on this @th-we @ahankinson ?

src/ExportGenerators.mss
```diff
@@ -984,6 +979,12 @@ function GenerateRest (bobj) {
         libmei.AddAttribute(r, 'color', nrest_color);
     }
 
+    position = bobj.RestPosition;
+    if (position != 0 and name != 'space')
+    }
+	libmei.AddAttribute(r, 'loc', (position + 64) / 16);
+    }
```
th-we, line 986: You have a wrong bracket and a wrong indentation here.

rettinghaus: I hate this. To make sure I have the correct indentation, I copied another line …

th-we:
I agree that the `@cue` attribute is horribly underdocumented.  [The only `@cue` sample](https://github.com/music-encoding/sample-encodings/blob/master/MEI_4.0/Music/Complete_examples/Parker-Gillespie_ShawNuff.mei#L209) uses it to mark notes as "cue size" notes, not as cues from another instrument (but one'could argue about that).  [Verovio agrees in the MusicXML import](https://github.com/rism-ch/verovio/blob/develop/src/iomusxml.cpp#L2164). MusicXML can distinguish if a note is actually a non-played note or just a small note, but Verovio treats them the same and uses `@cue` [for size rendering](https://github.com/rism-ch/verovio/blob/develop/src/iomusxml.cpp#L2382).

If `@cue` should actually be used for non-played cue notes from another instrument, we can't write it because Sibelius' `CueSize` is just a visual attribute that makes objects smaller.

In any case, @rettinghaus' suggestion to write more precise `@fontsize` attributes makes sense.

th-we:
Other places where `@cue` is only discussed/used for size:

* https://github.com/music-encoding/music-encoding/issues/509#issuecomment-384225530
* https://github.com/music-encoding/music-encoding/issues/558

But this is the discussion that lead to the addition of the `@cue` attribute, saying that it was meant to separate "small-ness" from "cue-ness":

* https://github.com/music-encoding/music-encoding/issues/433

Discussion related to the issue above:

* https://github.com/music-encoding/music-encoding/issues/406

I see a discrepancy between the attribute's apparent intent of its the actual usage.

th-we:
I [raised this as an issue](https://github.com/music-encoding/music-encoding/issues/668) on the schema repo.

rettinghaus:
Should I return from `@cue` to `@fontsize`?

annplaksin:
I remembered again why I chose `@fontsize`... because `@size="cue"` is transformed into `@fontsize="small"` in the mei30To40.xsl in the endoding tools, see https://github.com/music-encoding/encoding-tools/blob/64db706d578bf4fba6fd67158aa21cc768ac1ea6/mei30To40/mei30To40.xsl#L1969

Anyway, is there another chance that a note might be small or cuesized apart from being a cue note´?
Because we have the situation with sibmei, that we should not assume a logical interpretation for a visual feature when there is the chance of e.g. using using small notes for something other than their cue-ness...(to use that woderful word)
Since `@size="cue"` is a visual attribute, I would prefer staying in the visual domain and go to `@fontsize`... except it is 100% clear and totally obvious that this note must be a cue note.

th-we:
While there are small notehead styles to make heads smaller:

grafik

This does not affect accidentals or beams. To create something like cadenza-style runs of notes, there is to my knowledge no alternative to cue size notes:

grafik

(http://ks.imslp.info/files/imglnks/usimg/6/6e/IMSLP00016-Beethoven,_L.v._-_Piano_Sonata_16.pdf)

So yes, `CueSize` needs to be used in cases where the notes are not actually cues.

src/ExportGenerators.mss
```diff
@@ -973,17 +973,18 @@ function GenerateRest (bobj) {
...
     {
         nrest_color = ConvertColor(bobj);
         libmei.AddAttribute(r, 'color', nrest_color);
     }
 
+    position = bobj.RestPosition;
+    if (position != 0 and name != 'space')
+    {
+        libmei.AddAttribute(r, 'loc', (position + 64) / 16);
+    }
```
th-we, line 986: Unfortunately it's not that easy to determine the rest position. While I understand `@loc` as an absolute value (at least that's how Verovio handles it), `RestPosition` is relative to a default position. Sibelius' default position depends on if there are other voices and what is the content of those other voices. We'd have to reverse-engineer how Sibelius shifts rests.

Manual adjustments of rest positions is primarily of interest in multi-voice situations, but those are precisely the situations where we can't really tell what the rest position is.

Sorry for the hussle @rettinghaus , but could you remove this as well?

rettinghaus: Ok, didn't test it with multiple voices …

PR #143, opened 2020-05-27
Prepare Sibmei 4 release

th-we:
I'd love to release Sibmei 4 during MEC.  All tests are passing.

Next steps:

* Merge into master
* Create tag
* Create release
* Celebrate Sibmei4

Anything I forgot?


annplaksin:
Why did you delete the test runner?

th-we:
@annplaksin The original motivation for the test runner was that closing files after tests only seemed to work when running from the ribbon instead of the plugin editing dialog (more open windows makes Sibelius slower).  I just tested running with and without the test runner, but it did not seem to make a difference, so I removed it.  If you still like to use it, I'm happy to add it back.

plgconfig.js
```diff
@@ -5,9 +5,9 @@ var config = {
   // folder.
   plgPath:        './build',
   plgCategory:    'MEI Export',
-  pluginFilename: 'sibmei2.plg',
+  pluginFilename: 'sibmei4.plg',
   linkLibraries: [
-    'libmei.plg', 'sibmei_batch_mxml.plg', 'sibmei_batch_sib.plg', 'sibmei_test_runner.plg'
```
ahankinson, line 10: I don’t think libmei should be renamed here. Is there a reason for it?

annplaksin:
@th-we I am a lazy person, I like running all the test by clicking once in the menu. If you don't mind, I would love to keep the test runner... (even if closing the files afterwards would still be great)

ahankinson:
Yeah, if you’re going to remove it it should be a separate PR with its own discussion.

plgconfig.js
th-we: I renamed libmei to libmei4 so that sibmei4 can be installed side by side with sibmei2.  Like this, the user can generate both MEI 3 and MEI 4 if working with multiple MEI versions for different projects.  I'd suggest to make this approach standard for future major versions as well.

th-we: As libmei is specific to an MEI version, there would be conflicts if both Sibmei versions tried to access the same libmei.

AUTHORS
```diff
@@ -9,3 +9,5 @@ Initially developed with support from the Centre for Interdisciplinary Research
 Further enhancements and features were added with support from the National Endowment for the Humanities and Haverford College.
 
 Version 2.0 was developed as part of the Single Interface for Music Score Searching and Analysis project (https://simssa.ca), supported by the Social Sciences and Humanities Research Council of Canada, the Fonds de Recherche du Quebec – Société et Culture (FRQSC), and the Canadian Foundation for Innovation (CFI).
+
+Further work was sponsored by Österreichische Akademie der Wissenschaften, Max Weber Stiftung and Notengrafik Berlin GmbH.
```
rettinghaus, line 13: Please add Universität Leipzig

th-we:
@annplaksin Restored the test runner.

plgconfig.js
ahankinson: OK... I do need to point out that the libmei plugin has been [generated from libmei itself](https://github.com/DDMAL/libmei/blob/master/tools/langs/manuscript.py). I haven't had time to update it in a while, but some attention should be paid to backwards compatibility, or coming up with an alternate method of updating `libmei.plg` from the schema when new MEI releases appear.

ahankinson:
Is there a reason you are force-pushing everything? that's pretty dangerous

th-we:
@ahankinson I assume that as long as no one is doing any work on top of the pull request (which no one should until it is merged), there is no danger. If that's not the case, I'd better learn about what can go wrong.

plgconfig.js
th-we: @annplaksin updated libmei.plg to MEI 4 using libmei in 562e355.  I've also created some pull requests to libmei:

* https://github.com/DDMAL/libmei/pull/111
* https://github.com/DDMAL/libmei/pull/112
* https://github.com/DDMAL/libmei/pull/116

Can you explain your concerns regarding backwards compatibility?

ahankinson:
It's just strange development practice, since it effectively deletes what is on the server in favour of what you have locally. Generally you want git to warn you about these sorts of mismatches... Usually you only use force push when you know you've messed something up.

ahankinson:
Will wait for @annplaksin and @rettinghaus to approve this PR and then I will merge it.

plgconfig.js
ahankinson: No problem!

ahankinson:
Just waiting for @rettinghaus...

PR #145, opened 2020-05-28
Don't reference libmei4 in Initialize()

th-we:
libmei4.plg might not be loaded yet when Initialize() is run.

I didn't detect this problem when developing with unload/reload methods – it only occurs when re-starting Sibelius. This needs to be fixed before releasing v4.

I would be thankful if somebody else could run the tests as well to make sure everything is working smoothly not only on my machine. @annplaksin Could you do that?

### DIFF
--- a/plgconfig.js
+++ b/plgconfig.js
@@ -7,7 +7,7 @@ var config = {
   plgCategory:    'MEI Export',
   pluginFilename: 'sibmei4.plg',
   linkLibraries: [
-   'libmei4.plg', 'sibmei4_batch_mxml.plg', 'sibmei4_batch_sib.plg'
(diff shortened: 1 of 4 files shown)


annplaksin:
Sorry, it doesn't work on my machine :-(

And I'm sorry, but I am still concerned with my beloved test runner... it currently doesn't deploy with sibmei 4. I really like this thing for having a fast way to check if everything runs... (I know this might seem silly)

th-we:
@annplaksin That test helped a lot, thanks!  Does it work now?

annplaksin:
@th-we It worked and every test is passing on my machine.... and thanks for restoring the runner :-)

PR #153, opened 2020-06-15
Make sure we have an active score

th-we:
I randomly ran into a strange problem where indexing the score object was not ~~accessible~~ possible. This change is in the hope of giving a sensible message in case we for some obscure reason don't have an active score.

### DIFF
--- a/src/Run.mss
+++ b/src/Run.mss
@@ -45,13 +45,18 @@ function DoExport (filename) {
     Self._property:libmei = libmei4;
     libmei.destroy();
 
-    // Deal with the Progress GUI
     // set the active score here so we can refer to it throughout the plugin
     Self._property:ActiveScore = Sibelius.ActiveScore;
+    if (Self._property:ActiveScore = null)
+    {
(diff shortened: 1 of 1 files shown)


src/Run.mss
```diff
@@ -45,13 +45,18 @@ function DoExport (filename) {
     Self._property:libmei = libmei4;
     libmei.destroy();
 
-    // Deal with the Progress GUI
     // set the active score here so we can refer to it throughout the plugin
     Self._property:ActiveScore = Sibelius.ActiveScore;
+    if (Self._property:ActiveScore = null)
+    {
+        Sibelius.MessageBox('No score to convert. Can not export to ' & filename);
```
ahankinson, line 52: "Cannot"

ahankinson: It might help to say something more specific, though. Like "Could not find an active score. Unable to export to " & filename.

th-we: @ahankinson Thanks for checking – I followed your suggestions.

PR #154, opened 2020-06-15
Update libmei to fix "tail" bug

th-we:
See https://github.com/DDMAL/libmei/pull/117

@ahankinson Could you have a look at [my pull requests] on the libmei repo so we can get in synch again?

### DIFF
--- a/lib/libmei4.plg
+++ b/lib/libmei4.plg
@@ -2037,11 +2037,6 @@ GetTail "(element) {
         indent = indent - 1;
     }
 
-    if (Length(tl) > 0)
-    {
-        xmlout = xmlout & tabs & tl & Chr(10);
-    }
-
     // convertDictToXml takes care of adding the />
     // for tags that do not have children. We'll
     // take care of the terminal tag here for those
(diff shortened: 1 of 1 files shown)


ahankinson:
Is there an issue that describes the bug? I can’t really tell what triggers it...

th-we:
As `.SetTail()` is currently only used on a self-closing `<lb>` element., this is not triggered. But I'm experimenting with mixing text and elements, like:

    p = libmei.P();
    name = libmei.PersName();
    libmei.SetText(name, 'Joe');
    libmei.AddChild(p, name);
    name.SetTail(' sings');

Expected output:

    <p><persName>Joe</persName>sings</p>

Actual output is something like:

    <p><persName>Joe                 sings</persName></p>

(Handling of spaces when serializing will be another issue to tackle later.)

ahankinson:
Sounds OK. I wouldn’t expect anything to add extra spaces that were not present in the original.

th-we:
Thanks! Spaces are added because of indentation. Maybe we could add something like `libmei.SetIndent(element, false);` to suppress indentation for elements where it's undesired.

ahankinson:
If I see `name.SetTail(' sings');` then I would expect the output to be `<p><persName>Joe</persName> sings</p>` -- that the space in the original would be preserved in the XML output. Likewise, if there were no spaces, then I would not expect sibmei to inject a space when outputting elements, and the output (when rendered without tags) would be `Joesings`. That's the way HTML treats tags AFAIK. 

The spaces for the tail can probably be handled by removing the 'tabs' bit, which doesn't really make sense for a tail fragment.

PR #159, opened 2020-09-21
fix order in header

### LINKED ISSUE #158: encoding error in head
SIBMEI places the `encodingDesc` after the `workDesc`, which isn't allowed.

rettinghaus:
closes #158

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -13,6 +13,45 @@ function GenerateMEIHeader () {
     libmei.AddChild(header, fileD);
     libmei.AddChild(fileD, titleS);
 
+    respS = libmei.RespStmt();
+    libmei.AddChild(titleS, respS);
+    persN = libmei.PersName();
+    libmei.AddChild(respS, persN);
+    pubS = libmei.PubStmt();
+    libmei.AddChild(fileD, pubS);
+
+    avail = libmei.Availability();
+    ur = libmei.UseRestrict();
+    libmei.AddChild(avail, ur);
+    libmei.AddChild(pubS, avail);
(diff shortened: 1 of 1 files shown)


rettinghaus:
This has been fixed already in the develop branch, but it is also wrong in MEI3. 
So I suppose a last bugfix would be good here.

th-we:
Should we create a maintenance branch for v3 and apply this PR on that branch?  Naming suggestion: `mei3`. A little confusing because the current version is actually sibmei 2.3.0, but I think `mei3` is clearer than `sibmei2`.

rettinghaus:
Sounds like a good idea, but this PR goes to master branch, so I would merge it first and then make a new branch for this. 
Speaking of branches: shouldn't we rename master to main, and do we need a main branch at all, if we have dedicated branches for each version of MEI?

annplaksin:
One short question: Why would you like to still maintain the mei 3 version? I think the last release announced the coming upgrade to MEI 4. The Guidlines of v3 already deem the version outdated and refer to MEI 4... and lets see how long we have until the next MEI version will be released.

It's not that I want to spoil anything, I am just thinking about the effort to maintain two versions. Maybe it is better to stick with making a clean cut and finally release the version 4.
Just my two cents...

rettinghaus:
@annplaksin It's not we really maintain it, but we should make sure it always produces valid output. If something pops up elsewhere we could easily provide a bugfix release then.

bwbohl:
Although requested as reviewer, I think I'm not the right person to do this here… @th-we ?

bwbohl:
thx @th-we

Issue #152, opened 2020-06-14
Text subsitution variables


th-we:
[Currently, Sibmei assumes](https://github.com/music-encoding/sibmei/blob/c8c2155/src/ExportConverters.mss#L911) that substitution variables in text elements (such as `\$Composer\`) may only occur on their own, without surrounding text.  However, things like `Composed by \$Composer\ for her mum` are possible.

The substitution system is available in all text objects, including lyrics. Sibelius can give us all the text components with `.TextWithFormatting` (except for lyrics).  The result is an old-style TreeNode Array:

```
{
	 "Composed by "
	 "\B\"
	 "\$Composer\"
	 "\b\"
	 " for her mum"
}
```

th-we:
(As I'm working on text anyway, I'm taking care of this.)

th-we:
Works now (see [test file](https://github.com/music-encoding/sibmei/blob/develop/test/sibmeiTestSibs/text.sib)).

Issue #89, opened 2018-05-06
Fermatas over rests get exported twice


karend27:
When fermatas are placed over rests in Sibelius, they are encoded twice in the MEI output, both as an attribute within the rest element, and separately as a fermata element. This doesn't happen when fermatas are placed above notes, which are only encoded once as a separate fermata element.

To be consistent with how notes are treated, can fermatas above rests only be encoded as a fermata element, please.

```
<measure xml:id="m-1176" n="40">
							<staff xml:id="m-1177" n="1">
								<layer xml:id="m-1178" n="1">
									<note xml:id="m-1179" dur="breve" dur.ges="2048p" layer="1" oct="4" oct.ges="3" pname="d" pnum="50" staff="1" stem.dir="down">
										<verse xml:id="m-1187" n="1">
											<syl xml:id="m-1188" wordpos="t">e</syl>
										</verse>
									</note>
									<note xml:id="m-1180" dur="1" dur.ges="1024p" layer="1" oct="4" oct.ges="3" pname="e" pnum="52" staff="1" stem.dir="down">
										<verse xml:id="m-1189" n="1">
											<syl xml:id="m-1190" wordpos="i">la</syl>
										</verse>
									</note>
									<note xml:id="m-1181" artic="stacc" dur="1" dur.ges="1024p" layer="1" oct="4" oct.ges="3" pname="g" pnum="55" staff="1" stem.dir="down">
										<verse xml:id="m-1191" n="1">
											<syl xml:id="m-1192" wordpos="i">ue</syl>
										</verse>
									</note>
									<note xml:id="m-1182" dur="1" dur.ges="1024p" layer="1" oct="4" oct.ges="3" pname="a" pnum="57" staff="1" stem.dir="down">
										<verse xml:id="m-1193" n="1">
											<syl xml:id="m-1194" wordpos="m">ri</syl>
										</verse>
									</note>
									<note xml:id="m-1184" artic="stacc" dur="1" dur.ges="1024p" layer="1" oct="4" oct.ges="3" pname="a" pnum="57" staff="1" stem.dir="down">
										<verse xml:id="m-1195" n="1">
											<syl xml:id="m-1196" wordpos="t">te</syl>
										</verse>
									</note>
								</layer>
							</staff>
							<staff xml:id="m-1197" n="2">
								<layer xml:id="m-1198" n="1">
[...]

PR #162, opened 2021-02-05
Extension plugin support

th-we:
@annplaksin and I worked on how to export custom symbols and text styles as sensible MEI, without having to fork Sibmei. For this, we reworked the entire text and symbol export. This brings quite some substantial changes to the code, and the following benefits:

* Better conversion of formatting for all texts
* text substitution variables work everywhere (see #152)
* large `switch`es were replaced by declarative definitions of symbols
* fermatas were improved while porting them to the declarative approach (#89)

We wrote [documentation](https://github.com/notengrafik/sibmei/blob/f7fc603d502d02d6a4f9b1e16e0f038e69b24626/Extensions.md) and an [example](https://github.com/notengrafik/sibmei/blob/f7fc603d502d02d6a4f9b1e16e0f038e69b24626/lib/sibmei4_extension_test.plg).


src/ExportConverters.mss
```diff
@@ -592,15 +592,11 @@ function ConvertSibeliusStructure (score) {
     {
         for each Bar b in s
         {
-            if (bar_to_staff.PropertyExists(b.BarNumber))
```
ahankinson, line 595: Are these changes part of the plugin support?

th-we: There are a few changes not strictly related to text, symbols and extensions that I made while working with the code.  That's the case here and in both the Run.mss files.  Some things are only for clarity and cosmetics (like here code that does the same thing, but a bit clearer; or inconsistent indentation in one Run.mss).  Some things are for easier development and use, like better file name defaults when exporting or making it possible to automatically calling the mocha tests after exporting the test set.

annplaksin:
Are there any objections about this PR? If not, I would like to merge it maybe next week or so...

ahankinson:
I can merge it

annplaksin:
Thanks a lot!

ahankinson:
THank you (and @th-we ) for the work!

PR #167, opened 2021-04-30
New: support spiccato articulations

mss2221:
This commit adds support for spiccato. Sibelius calls this a wedge, but in MusicXML called spiccato. Both note properties and symbol variant are converted. Note property added as attribute of note element, and symbol added as child element.

http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#EL-MusicXML-spiccato.htm%3FTocPath%3DMusicXML%2520Reference%7CScore%2520Schema%2520(XSD)%7CElements%7Cnote%7C_____96


th-we:
Thanks!

PR #168, opened 2021-04-30
Fixed: expose build as a gulp task

mss2221:
This commit exposes the build function as a separate gulp task allowing the user to run gulp build independent of develop.


th-we:
Makes sense. Thanks!

PR #172, opened 2021-05-09
New: support articulations symbols above and below note

mss2221:
This commit adds support for symbol variants of staccato, staccatissimo, spiccato and marcato below the note, and for staccato and staccatissimo above.


rettinghaus (approved):

rettinghaus:
thanks!

PR #176, opened 2021-05-17
Avoid added whitespace around text

th-we:
Originally was part of #173

### DIFF
--- a/test/mocha/test-text.js
+++ b/test/mocha/test-text.js
@@ -43,6 +43,11 @@ describe("Text elements", function() {
         const subscript = xpath.evaluateXPath("//*:measure[@n='1']//*:title[@type='subordinate']/*:rend[@rend='sub']", meiText);
         assert.notStrictEqual(subscript.length, 0, "Subscript in subtitle is missing");
     });
+    // Formatting elements must be output without indentation or any other additional whitespace
+    it("check for whitespace introduced by formatting tags", function() {
+        const subtitle = xpath.evaluateXPath("//*:measure[@n='1']//*:title[@type='subordinate']", meiText);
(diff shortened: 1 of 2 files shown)


test/mocha/test-text.js
```diff
@@ -43,6 +43,11 @@ describe("Text elements", function() {
         const subscript = xpath.evaluateXPath("//*:measure[@n='1']//*:title[@type='subordinate']/*:rend[@rend='sub']", meiText);
         assert.notStrictEqual(subscript.length, 0, "Subscript in subtitle is missing");
     });
+    // Formatting elements must be output without indentation or any other additional whitespace
+    it("check for whitespace introduced by formatting tags", function() {
+        const subtitle = xpath.evaluateXPath("//*:measure[@n='1']//*:title[@type='subordinate']", meiText);
+        assert.strictEqual(subtitle.textContent, 'E0=mc2', 0, "Text must not have any whitespace");
```
ahankinson, line 49: I'm not sure I'm reading this correctly, but just to check: XML processors should interpret line endings as white space, not nothing, as far as I know. In HTML something like this:

```
<my-tag>
 this
 is
 some
 text
</my-tag>
```

Should result in "this is some text", not "thisissometext". 

Is that behaviour still preserved here? Or is it changed somehow?

th-we: Only the whitespace that was in the original text in Sibelius is written out. This means no added line breaks and indentation. Do you wish for more line breaks and indentation?

The point of this pull request is that this example:

grafik

would previously be exported like "E <sub>0</sub> =mc <sup>2</sup>" (added whitespace resulting from indentation slightly exaggerated for illustration).

ahankinson: OK, fine -- I just wanted to check.

PR #178, opened 2021-05-17
Libmei SetId() and RemoveChild() fix

th-we:
... to not break getElementById() or getElementsByName()

### DIFF
--- a/lib/libmei4.plg
+++ b/lib/libmei4.plg
@@ -1749,6 +1749,9 @@ AddChild "(element, child) {
         cid = child._id;
         child._parent = element._id;
         element.children.Push(cid);
+        // The following might be redundant, but in case a child is removed from
+        // one parent and added to another, it's safer to re-register the ID.
+        Self.MEIFlattened[cid] = child;
 }"
 RemoveChild "(element, child) {
     child._parent = null;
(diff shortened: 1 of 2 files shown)


ahankinson:
`removeKeyFromDictionary` does what it says -- it removes the key from the dictionary. I would expect that the behaviour would differ when calling `PropertyExists` on a dictionary with the ID, but an empty value?

th-we:
As `MEIFlattened` becomes quite large, I felt it was not a good idea to re-create it every time `SetId()` is used.  But at the moment, Sibmei only uses `SetId()` in the header and it's therefore not an issue.  If it was used in the body, that might mean O(n²) performance.

I also realize that one change required for this slipped into the wrong commit.

https://github.com/music-encoding/sibmei/pull/173/commits/e9c576982152ceef8c93fe50db96451b250c86c0#diff-8f908fa55702e2b03b5c2472511fd19030738d0a0f501b61089457bfb87f6354R1887

I can either include that missed line, or rework the pull request to use the old approach of re-generating the dictionary.

ahankinson:
There's something just a bit icky about keeping IDs around that are 'deleted' -- I would probably prioritise consistency and correctness over speed. What do you think?

(Of course in a 'real' language we wouldn't have these problems, but...)

th-we:
`SetId()` isn't so much of a problem, but `RemoveChild()` might be because we do use it in the body.

At the moment that will not make an impact as the string concatenation happening in `convertDictToXml` is the bottleneck for larger files.  But that can be changed easily.

ahankinson:
But only in the tuplet handling code, right? That's probably unlikely to have a huge speed impact.

(You're a brave man, sticking your fingers in the tuplet code. Thar be dragons. )

th-we:
I don't deserve the credit, you wrangled the dragons!

https://github.com/music-encoding/sibmei/blame/develop/src/ExportProcessors.mss#L168-L183

I'll modify the pull request and restore the original behavior to clone the Dictionary.

th-we:
@ahankinson  Any objections against merging the version with `removeKeyFromDictionary()` reverted to its old behavior?  The removed line in the tuplet code is because I moved that line to `RemoveChild()`.  `RemoveChild()` should do all the necessary bookkeeping to disassociate parent and child.

lib/libmei4.plg
```diff
@@ -1814,14 +1819,15 @@ SetAttributes "(element, new_attrs) {
...
-SetId "(element, value) {
-        olddict = Self._property:MEIFlattened;
-        oldid = element._id;
-
-        newdict = removeKeyFromDictionary(olddict, oldid);
-        element._id = value;
-        newdict[value] = element;
-        Self._property:MEIFlattened = newdict;
+SetId "(element, newId) {
+    UnregisterId(element._id);
```
ahankinson, line 1823: Isn’t this supposed to be `newId`? Where does `value` come from?

ahankinson: Sorry — next line. Mobile phone code reviews :-/

th-we: That's right. I made sure I unloaded/reloaded the proper files in Sibelius and ran the tests successfully.

PR #179, opened 2021-05-17
Limit progress update frequency for performance

th-we:
Applying an "old trick" of ManuScript development:

http://sibelius-manuscript-plug-in-developers.3224780.n2.nabble.com/Performance-notes-for-Plug-in-Developers-td7573341.html

Depending on the kind of score, this can in the most extreme cases (sparsely filled score with a single staff) cut down processing time by about 40%.

### DIFF
--- a/src/ExportGenerators.mss
+++ b/src/ExportGenerators.mss
@@ -211,19 +211,26 @@ function GenerateMEIMusic () {
 
     currentScoreDef = null;
 
+    timerId = 1;
+    Sibelius.ResetStopWatch(timerId);
+
+
     for j = 1 to numbars + 1
     {
-        // this value may get changed by the volta processor
-        // to inject a new parent in the hierarchy.
-        progressMsg = utils.Format(_ExportingBars, j, numbars);
-        cont = Sibelius.UpdateProgressDialog(j, progressMsg);
-        section = Self._property:SectionElement;
-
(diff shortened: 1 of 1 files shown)


src/ExportGenerators.mss
```diff
@@ -211,19 +211,26 @@ function GenerateMEIMusic () {
...
+        // http://sibelius-manuscript-plug-in-developers.3224780.n2.nabble.com/Performance-notes-for-Plug-in-Developers-td7573341.html
+        if (Sibelius.GetElapsedMilliSeconds(timerId) > 200)
         {
-            ExitPlugin();
+            Sibelius.ResetStopWatch(timerId);
+            progressMsg = utils.Format(_ExportingBars, j, numbars);
+            if (not Sibelius.UpdateProgressDialog(j, progressMsg))
+            {
+                // if the user has clicked cancel, stop the plugin.
+                ExitPlugin();
```
ahankinson, line 230: Will this not ignore a 'Cancel' click for five seconds?

th-we: It may only react after a bit over 0.2 seconds. But the cancel click will be registered in any case. I just counted >20 uses of this trick in plugins we have in regular use.  It's proven to work and can make a big difference.

ahankinson: OK

PR #181, opened 2021-09-01
Make sure measure/@n gets actually written

th-we:
Because of ManuScript type coercion, `@label` was only written when ExternalBarNumber was the string '1'. Some fun facts about ManuScript:

* `not not '1'` => false (character `'1'`)
* `not not ('1' & '')` => true (string `'1'`)
* `not not ('t')` => true (character `'t'`)
* `not not ('t' & '')` => false (string `'t'`)

Also, `ExternalBarNumberString` is always set, not only if we have explicitly set deviating bar numbers.

Alternative would be to always write `@label`, independent of whether it differs from `@n`.


rettinghaus (approved):

ahankinson (approved):

ahankinson:
Ready to merge?

th-we:
Yes, ready. As you both approved, I'll merge myself.

ahankinson:
No worries -- I was just about to do it anyway!

PR #189, opened 2021-09-28
Move filename dialog to DoExport() and return filename

th-we:
Like this, project specific export plugins easily use the standard sibmei behavior for determining file names. The returned file name can also be used to trigger postprocessing for the exported file.

### DIFF
--- a/src/Run.mss
+++ b/src/Run.mss
@@ -14,6 +14,12 @@ function Run() {
         return False;
     }
 
+    DoExport(null);
+
+}  //$end
+
+
+function GetExportFileName () {
     // get the active score object
     activeScore = Sibelius.ActiveScore;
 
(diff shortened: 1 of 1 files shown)


src/Run.mss
```diff
@@ -14,6 +14,12 @@ function Run() {
         return False;
     }
 
+    DoExport(null);
```
ahankinson, line 17: Should this be `filename = DoExport(null);` since `DoExport` returns the filename?

th-we: ManuScript is fine with dropping unused return values.

ahankinson: I guess I can rephrase the question: I don't understand why you're making the change if you're ultimately not going to use the value?

th-we: Ah, I understand.  `DoExport()` serves as the entry point to sibmei for other plugins calling sibmei. Specifically, I call sibmei from the CMO export plugin here:

https://github.com/notengrafik/cmo_sibmei/blob/18b2757c48eeb5bffd57a591ec01fee2b0c28bf7/src/Run.mss#L31

Instead of hard-coding how the export file name is generated or re-implementing the behavior of the file name dialog, I could just use `DoExport()` to determine the filename and then continue working with it.

ahankinson: That makes sense; thanks. I guess it doesn't really make sense to return a value from `Run`?

th-we: I don't think so. But it wouldn't do any harm either. But thinking about this pull request further, it might make sense to move more functionality from `Run()` to `DoExport()` – specifically the version test and the ScoreCount because they make sense no matter if sibmei is run standalone or called by another plugin. I'll add another commit.

PR #199, opened 2022-10-02
improve tuplet attributes, rebased

th-we:
Rebased #137 on top of develop plus tests.


rettinghaus:
@th-we could you please add a screenshot of the test file here?

th-we:
Looks like this:

grafik

(The ratio style in the last bar is not supported by MEI and is therefore ignored at the moment.)

rettinghaus (approved):
Thanks for adding these tests!

PR #201, opened 2022-10-02
add technique text handling

rettinghaus:
This adds a `CreateDirective` function to handle staff text, in this case `text.staff.technique`. 
Also it unifies some names.


rettinghaus:
@th-we Could you please have another look here?

th-we:
@rettinghaus Can you merge my new pull request on your branch, and then we merge this pull request here?

rettinghaus:
Done. Sorry I overlooked it.

src/TextHandler.mss
```diff
@@ -111,27 +112,35 @@ function FiguredBassTextHandler (this, textObject) {
...
-    AddFormattedText(anchoredText, textObj);
+    AddFormattedText(anchoredText, textObject);
     return anchoredText;
 }  //$end
 
+function CreateDirective (this, textObject) {
+    directive = GenerateControlEvent(textObject, 'Dir');
+    AddFormattedText(directive, textObject);
+    styleIdPrefix = 'text.staff.';
+    text_style = Substring(textObject.StyleId, Length(styleIdPrefix));
```
th-we, line 125: @rettinghaus It's clear that for what you implemented, we'll always get `@class='technique'`. I think we should document that in this form, this function is only to be registered as a text handler for `StyleId`, not for `StyleAsText` because for user defined styles, we would get not so helpful class names, like `text.staff.technique.user.0000123`. Or we can extend this a little and check if `MSplitString(o.StyleId, '.')[-2] = 'user'`, and in that case, use `textObject.StyleAsText` for the class attribute. What do you think?

rettinghaus: Are user defined styles a common thing? Then we definitely do the latter.

rettinghaus: Added a check for user defined styles.

th-we:
@rettinghaus I created another pull request to fix an error.

src/TextHandler.mss
th-we: > Are user defined styles a common thing?

At least in our work it is.

src/TextHandler.mss
```diff
@@ -111,27 +112,39 @@ function FiguredBassTextHandler (this, textObject) {
...
+    if (MSplitString(o.StyleId, '.')[-2] = 'user')
+    {
+        text_style = textObject.StyleAsText;
+    }
+    libmei.AddAttribute(directive, 'class', text_style);
```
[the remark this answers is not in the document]
th-we: Ah, it should actually be `@type`, right?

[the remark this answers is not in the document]
ahankinson: `@type` would allow it, but there's no formalized specification for what the value of "technique" means. (You could also call it `@type="dinosaur"` and it would have the same level of interoperability.)

src/TextHandler.mss
```diff
@@ -111,27 +112,39 @@ function FiguredBassTextHandler (this, textObject) {
...
-    AddFormattedText(anchoredText, textObj);
+    AddFormattedText(anchoredText, textObject);
     return anchoredText;
 }  //$end
 
+function CreateDirective (this, textObject) {
+    directive = GenerateControlEvent(textObject, 'Dir');
+    AddFormattedText(directive, textObject);
+    styleIdPrefix = 'text.staff.';
+    text_style = Substring(textObject.StyleId, Length(styleIdPrefix));
```
ahankinson, line 125: This should probably be `textStyle` since camelCase is generally the style here.

src/TextHandler.mss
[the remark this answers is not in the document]
rettinghaus: Having a URI pointing nowhere isn't very nice. I'd prefer `@type`, but if there are spaces in `StyleAsText` that would also be bad. For the time being I switched to `@label` as only attribute allowing plain text.

ahankinson:
I'm going to approve this.

PR #205, opened 2023-02-13
Fixed: Use corresp instead of startid for fb

### LINKED ISSUE #192: Figured bass figures are not attached to the correct point
From rism-digital/verovio#2533

> The problem seems to come from the fact that all harm/fb are having a `@startid="#m-121"`, which means they are all positioned under the same note. Also, there are some empty fb, which do not make sense. Removing them and using only tstamp without `@startid` fixes the issue

Tagging @vivianteresa for visibility on this.

ahankinson:
Changes the `<harm>` element for figured bass to use `@corresp` instead of `@tstamp` for the note that it is attached to. This should still correspond to the nearest previous bar object. This is to deal with situations like the following, where the figure is attached to the first note in the bar, but should be positioned on the third beat.

image

See the discussion in #192 for additional context.

Fixes #192

### DIFF
--- a/src/TextHandler.mss
+++ b/src/TextHandler.mss
@@ -104,6 +104,15 @@ function TempoTextHandler (this, textObject) {
 function FiguredBassTextHandler (this, textObject) {
     // 'text.staff.space.figuredbass'
     harm = GenerateControlEvent(textObject, 'Harm');
+
+    // uniquely, for figured bass we do not use the startid here,
+    // since a figure can change halfway through a note. So we remove
+    // the startid and replace it with corresp, pointing to the
+    // same ID.
+    startidValue = libmei.GetAttribute(harm, 'startid');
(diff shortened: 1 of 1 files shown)


rettinghaus (approved):
I'm not super happy to use `@corresp` for this, but it's still the best move we can make here, I guess.

ahankinson:
Thanks! I verified with Laurent that it would not affect Verovio to use `@corresp` but that it would fix the specific issue: https://github.com/rism-digital/verovio/issues/2533

craigsapp:
> I'm not super happy to use `@corresp` for this, but it's still the best move we can make here, I guess.

It seems like an interesting idea and would be useful if you need to realize the figured bass since the `@corresp` would help you find the note that goes with the figures.

Similar to the finding the duration/pitches of the repeated chord symbol discussed last week on the mei-encoding repo (I think).

PR #208, opened 2023-02-13
Fixed: Dialog layout

ahankinson:
Fixes some issues with buttons being too big and too close together, so they were cutting off text.


ahankinson:
If someone can merge this, I think it's the last blocker to a release. Thanks!

rettinghaus (approved):
Haven't tested this, but looks good to me.

PR #211, opened 2023-09-10
Export colored lyrics

th-we:
Just a tiny commit, but I will hopefully be able to do more Sibmei work in the coming months.


rettinghaus (approved):
Thank you!

PR #212, opened 2023-09-22
Color for control events

th-we:
Also fix one control event test. (Code was changed from `class` attribute to `label` attribute, but test wasn't.)


rettinghaus (approved):
Great, thank you!

PR #213, opened 2023-09-22
Unify and improve batch export code

th-we:
Code for batch export from Sibelius and MusicXML was almost identical, but slowly diverged. Now it's unified again.

Additional improvements:
* Three MusicXML extensions are exported: *.xml, *.musicxml and *.mxl
* Hopefully more useful error message (also for non-batch export) if file could not be written
* Final message added
* Files are processed in a predictable order (alphabetical). The order of files when looping over a folder is otherwise rather random.

The progress dialog was removed because after a split second, it is overridden by the progress dialog looping over the bars of each file anyways.

Could someone please test on a Mac before merging?  Especially how the MusicXML export works.

### DIFF
--- a/lib/sibmei4_batch_mxml.plg
+++ b/lib/sibmei4_batch_mxml.plg
@@ -8,44 +8,28 @@
     Run "()
     {
         folder = Sibelius.SelectFolder();
-
-        if (IsObject(folder))
+        if (null = folder)
         {
-            sibmei4.InitGlobals(null);
-
-            // count files for progress dialog
-            numFiles = folder.FileCount('XML');
-            index = 0;
-            Sibelius.CreateProgressDialog(_Processing, 0, numFiles);
-
-            for each XML file in folder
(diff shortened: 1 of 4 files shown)


rettinghaus:
With the latest Sibelius version I get this: 
image

th-we:
Huh? That is really strange.  Can you check if `sibmei4.plg` is somewhere in your Sibelius Plugins folder and that it was not accidentally unloaded?

Thanks a bunch for testing, @rettinghaus!

th-we:
@rettinghaus If you could try again that would be superb. I pushed two more commits, which are most likely unrelated to the problem you reported, but anyways...

rettinghaus:
So I did more tests. The first problem was indeed my fault, but now I'm getting this error

image

I tried different folders and also the Batch-Sibelius-Export, all with the same result. 

Running Sibelius 2023.8

PR #215, opened 2023-09-29
Add color support for multiple elements

rettinghaus:
This PR adds export of color of

* clefs
* fermatas
* slurs/trills/octaves/lines (all Sibelius' 'line' elements)


th-we:
I'll be offline for the next week and don't have time to try this right now – unless someone else has a look, I'll check this out the week after next week.

PR #218, opened 2023-11-29
Make text substitution case-insensitive

th-we:
Turns out that the text substitution syntax is case-insensitive, i.e. we can use any of `\$Title\`, `\$TITLE\`, `\$title\`, `\$TiTlE\` and so on. I updated the test file to use different variants.


rettinghaus (approved):
looks good to me

rettinghaus:
@ahankinson please have a look and merge.

ahankinson:
It looks fine, but I can’t really find a description of how it’s supposed to be used?

th-we:
@ahankinson Do you mean how to use the substitutions in Sibelius? If you edit this text block in the [test file](https://github.com/music-encoding/sibmei/blob/dc73b4bef1452b67db65c3c07ded248bba5fd228/test/sibmeiTestSibs/text.sib):

grafik

the editable text looks like this:

grafik

The actual `Text` and `TextWithFormattingAsString` properties do not include the part after the "=" sign, they look like this:

```
Text: Bold \$TITLE\ Italic \$Subtitle\ underlined \$dedication\
TextWithFormattingAsString: Bold \B\\$TITLE\\n\\b\Italic \I\\$Subtitle\\n\\i\\U\underlined \$dedication\
```
Editing the part after "=" will actually modify the referenced metadata propery.

Not sure if this answers your question.

th-we:
Anything I should do on this PR?

PR #219, opened 2023-12-05
Support elision syllables separated by spaces

th-we:
In situations like this:

grafik

...encodes the lyrics under the second note as:
```
<verse n="1">
    <syl wordpos="t">re</syl>
    <syl con="d" wordpos="i">au</syl>
</verse>
```


src/Utilities.mss
```diff
@@ -1,41 +1,59 @@
 function MSplitString (string, delimiter) {
     //$module(Utilities.mss)
     /*
-        The default Splitstring method is buggy,
-        so I've re-implemented it here.
+        The default SplitString method returns a TreeNode Array which is hard
+        to handle and can lead to Sibelius crashes.  This is a more friendly
```
rettinghaus, line 5: Double space.

src/Utilities.mss
```diff
@@ -1,41 +1,59 @@
...
+
+    // In the loop, we push the component and the following delimiter, but
+    // there is no delimiter following the last component, so we remove the last
+    // item again (which is always an empty string).
+    ret.Pop();
+
     return ret;
 }  //$end
 
+
```
rettinghaus, line 56: Why add an empty line?

src/ExportProcessors.mss
```diff
@@ -297,16 +297,21 @@ function ProcessLyric (lyricobj, objectPositions) {
...
         libmei.SetText(sylel, syllables[0]);
 
         // Handle any found elisions
-        for s = 1 to syllables.Length
+        for s = 1 to syllables.Length step 2
         {
-            libmei.AddAttribute(sylel_last, 'con', 'b');
+            elisionDelimiter = syllables[s];
+            elisionSyl = syllables[s + 1];
+            if (elisionDelimiter = '_')
```
rettinghaus, line 309: If there is a space between syllables it is preferred to also encode this with `@con="s"`, see https://music-encoding.org/guidelines/v4/attribute-classes/att.syl.log.html

rettinghaus (changes requested):
Looks good. Some minor comments.

src/Utilities.mss
th-we: A habit of mine to add double spaces after full stops. I can change it.

src/Utilities.mss
th-we: In the diff it looks like I added a single empty line, but the `return ret;` originally belonged to `MSplitString()` that I changed.  The new `return ret;` belongs to `SplitStringIncludeDelimiters()`.  Shall I change it?

Should we create a style guide?  In the codebase, we have one and two spaces between functions.

th-we: That's what I prefer as well and what I did here.

rettinghaus (approved):

PR #220, opened 2023-12-12
Fix and extend meter export

th-we:
grafik

Changes to previous behavior for this example:

* Hides time signature in bar 1
* Writes a time signature for bar 2 (even though it doesn't change, but it's repeated)
* Writes "cut" symbol for bar 5 (was not recognized as such before)
* Writes a time signature for bars 5 and 6 (even though it remains 2/2)
* Exports compound time signatures in bar 7


rettinghaus (approved):
Looks really good, thank you!

PR #221, opened 2023-12-12
Prevent exporting an empty <mdiv> if last bar is a section end

th-we:
If the last bar looks like this:

grafik

... there used to be an empty `<mdiv>` at the end.


rettinghaus (approved):
Thanks for finding and fixing!

PR #222, opened 2024-02-09
Rework text, symbol and line handlers

th-we:
This pull requests makes handling of text, symbols and lines work the same way.  Specifically, we can get rid of the [double bookkeeping](https://github.com/music-encoding/sibmei/blob/90849923ed792c807c89e707c80745192d0b8575/src/SymbolHandler.mss#L14-L112) for symbols, and we can use templates for text objects now, [adding text in a declarative way](https://github.com/notengrafik/sibmei/blob/a864f1b920109271ba1e34f4fd006cac09873994/src/TextHandler.mss#L9-L26).  This also should make it simpler to create extensions, even for people with very limited coding experience.

The declarative text feature made a breaking change in the API necessary because `MeiFactory()` needed a second argument.  I took this as a chance to rework the API and make it easier to document and (hopefully) to use and maintain.  It now works with fewer methods.  Additionally, I changed all the related "internal" code to work the same way and use the same names, so documentation applies to both extensions and internal code and [should be pretty thorough now](https://github.com/notengrafik/sibmei/blob/a864f1b920109271ba1e34f4fd006cac09873994/ExportHandlers.md).

Old extensions should still work - they are passed a [legacy version of the API](https://github.com/notengrafik/sibmei/blob/a864f1b920109271ba1e34f4fd006cac09873994/src/Extensions.mss#L186-L198).  (Would be interesting to know if there actually are any such extensions out there apart from the 2-3 I've written myself.)


plgconfig.js
```diff
@@ -7,7 +7,7 @@ var config = {
   plgCategory:    'MEI Export',
   pluginFilename: 'sibmei4.plg',
   linkLibraries: [
-    'libmei4.plg', 'sibmei4_batch_mxml.plg', 'sibmei4_batch_sib.plg', 'sibmei4_test_runner.plg', 'sibmei4_extension_test.plg'
+    'libmei4.plg', 'sibmei4_batch_mxml.plg', 'sibmei4_batch_sib.plg', 'sibmei4_test_runner.plg', 'sibmei4_extension_test.plg', 'sibmei4_legacy_extension_api_v1_test.plg'
```
ahankinson, line 10: I’m really not a huge fan of copying these examples to the installation by default. Is this necessary? It just seems to add a lot of plugins that most users will never use.

th-we: This is necessary during development because when working on the extensions, we would otherwise have to copy&paste them manually.

But I agree it was a mistake to include the test code in the last release.  That should be avoided.

ahankinson: Can you just change it locally and not commit your changes to the config?

th-we: Everybody running the tests would have to do that.  Otherwise the tests will crash because of missing plugins.

ahankinson: I think that’s something that can be documented, no? The only people running the tests are devs, and  it would clean up the UI for everyone else.

th-we: But everybody else should probably use a release version where we only package the necessary files (which we didn't do last time).  If we can avoid working with a permanently "dirty" working directory, I'd highly welcome that.  If we stash something, checkout another branch, continue working, then we end up with the wrong version of the test plugins.  We have to remember to change the file every time we switch a branch.  Or changed files will accidentally not be updated, we test with the wrong version of the files and don't notice they are broken.

For our inhouse plugins, our own compiler creates test and release versions of the plugins.  That could be possible for Sibmei as well, but I don't think that's worth the effort.

ExportHandlers.md
```diff
@@ -0,0 +1,253 @@
+# Handling of Text, Lines and Symbols
+
+Text and Lines have in common that there are a multitude of different styles, and users can add their own styles.  The style of a text or line object can be determined by looking at their `StyleId` and/or `StyleAsText` properties.  Each style should be exported in a different fashion.
```
rettinghaus, line 3: I know it's a habit of yours to put a double space after a period. I don't care too much, but at least it should be consistent. In some cases there is only a single space.

ExportHandlers.md
```diff
@@ -0,0 +1,253 @@
...
+api.RegisterTextHandlers('StyleId', CreateDictionary(
+    'text.staff.technique', CreateSparseArray(
+        'Dir',
+        CreateDictionary('label', 'technique'),
+        api.FormattedText
+    )
+), Self);
+```
+
+Templates declaratively describe an MEI element by means of ManuScript data structures  for example:
```
rettinghaus, line 73: This one's definitely not good:
```suggestion
Templates declaratively describe an MEI element by means of ManuScript data structures for example:
```

ExportHandlers.md
```diff
@@ -0,0 +1,253 @@
...
+    )
+)
+```
+
+The placeholder will be replaced by an ID reference when writing the XML. Which ID is written depends on the line's end position and the value of the placeholder:
+
+* `'PreciseMatch'`: `@endid` will only be written if there is a NoteRest precisely at the `EndPosition` in the same voice as the line.
+* `'Next'`: If there is no NoteRest at the `EndPosition`, will write an `@endid` pointing to the closest following NoteRest, if there is one in the same voice as the line.
+* `'Previous'`:  If there is no NoteRest at the `EndPosition`, will write an `@endid` pointing to the closest preceding NoteRest, if there is one in the same voice as the line.
+* `'Closest'`: Writes an `@endid` that points to the closest NoteRest at the `EndPosition` in the same voice as the line.
```
rettinghaus, line 169: Why not mark "NoteRest" as well?

ExportHandlers.md
```diff
@@ -0,0 +1,253 @@
...
+
+* A [template](#registering-a-template)
+* The text, line or symbol object for which an element is generated by means of the template.  This parameter is only of importance if [dynamic template fields](#dynamic-template-fields) are used.
+
+#### `AddFormattedText()`
+
+Takes two arguments:
+
+* `parentElement`: MEI element that the formatted text nodes should be appended to.
+* `textObj`: A  `Text` or `SystemTextItem` object. Its `TextWithFormatting` property is converted to MEI markup.
```
rettinghaus, line 245:
```suggestion
* `textObj`: A `Text` or `SystemTextItem` object. Its `TextWithFormatting` property is converted to MEI markup.
```

rettinghaus (commented):
Looks really good to me. Thank you for the extensive documentation. 
Some minor remarks from my side.

ExportHandlers.md
th-we: O.K., why not!

rettinghaus (approved):

PR #225, opened 2024-02-29
Prevent crash when frontmatter has formatted text

th-we:
Also updated test file to include formatted frontmatter


th-we:
This will cause conflicts with #224, so I consider adding this commit with the updated test file to #224, closing this commit again.

th-we:
I added this commit to #224 now.

PR #227, opened 2024-03-14
Introduction of lyric handlers

th-we:
Another big pull request.

In analogy to line, text and symbol handlers, this introduces lyrics handlers. Lyrics are a little special and made some adjustments to the templating system necessary.  Namely, I introduced what I called "[Template Actions](https://github.com/notengrafik/sibmei/blob/pr/lyric-handler/ExportHandlers.md#internals-template-actions)".

I ditched compatibility with legacy API version 1 because it's much easier to upgrade extensions than to maintain backwards compatibility.

Now I think the only missing thing where users can define their own styles is noteheads.


src/ExportHandlers.mss
```diff
@@ -1,52 +1,128 @@
...
+        for each Name validIdProperty in handlers
+        {
+            validIdProperties.Push(validIdProperty);
+        }
+        Sibelius.MessageBox(
+            'Error in extension plugin \''
+            & pluginInfo._extensionInfo.plugin.Name
+            & '\': Expected either of \''
+            & validIdProperties.Join('\' or \'')
+            & ' as first paramter of registration function, but found \''
```
ahankinson, line 116:
```suggestion
            & ' as first parameter of registration function, but found \''
```

src/ExportHandlers.mss
```diff
@@ -1,52 +1,128 @@
...
+            & idProperty
+            & '\'.\n\nPlugin execution is aborted. To continue, deactivate or fix \''
```
ahankinson, line 118:
```suggestion
            & '\'.\n\nPlugin execution is stopped. To continue, deactivate or fix \''
```

src/TextHandler.mss
```diff
@@ -88,7 +95,7 @@ function FiguredBassTextHandler (this, textObject) {
 }  //$end
 
 
-function AddFormattedText (parentElement, textObject) {
+function AddFormattedText (self, parentElement, textObject) {
```
ahankinson, line 98: The addition of `self` here is a bit funny, since it's not used?

src/ExportConverters.mss
```diff
@@ -700,25 +700,6 @@ function ConvertPositionWithDurationToTimestamp (bobj) {
     return measureDuration & 'm+' & position;
 } //$end
 
-function ConvertTupletStyle (tupletStyle) {
```
ahankinson, line 703: This was removed, but I can't see where it was replaced?

th-we: This is just removal of unused code in commit 2430053674d41af1c482834d8602d4e87fea183f.

src/TextHandler.mss
th-we: This is because I moved to "Template Actions", and `AddFormattedText` now became the Template Action method of the `FormattedText` placeholder object. Methods always get a "self" as first argument, but it's of no use here.

src/ExportHandlers.mss
th-we: Many thanks for looking at this so promptly, Andrew. I think I'll wait if there will be more suggestions like this and create a big commit with cosmetic/wording changes.

src/LineHandler.mss
```diff
@@ -8,100 +8,100 @@ function InitLineHandlers () {
...
-        // 'line.staff.harmonic.pinch',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        // 'line.staff.harmonic.touch',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        // 'line.staff.harmonics',                CreateSparseArray('Line', CreateDictionary('form', 'solid')),
+        'line.staff.guitareffect',                  @Element('Line', @Attrs('form', 'dashed', 'type', 'guitareffect', 'endid', 'PreciseMatch')),
+        // 'line.staff.harmonic.artificial',           @Element('Line', @Attrs('form', 'solid')),
+        // 'line.staff.harmonic.harp',                 @Element('Line', @Attrs('form', 'solid')),
+        // 'line.staff.harmonic.pinch',                @Element('Line', @Attrs('form', 'solid')),
+        // 'line.staff.harmonic.touch',                @Element('Line', @Attrs('form', 'solid')),
+        // 'line.staff.harmonics',                     @Element('Line', @Attrs('form', 'solid')),
         // The Hauptstimme line type only has a start symbol and an end symbol
```
rettinghaus, line 45: This comment is obsolete now.

src/LineHandler.mss
```diff
@@ -8,100 +8,100 @@ function InitLineHandlers () {
...
-        // 'line.staff.string.below.2',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        // 'line.staff.string.below.3',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        // 'line.staff.string.below.4',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        // 'line.staff.string.below.5',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        // 'line.staff.string.below.6',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        // 'line.staff.string.below.7',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        // 'line.staff.string.below.8',           CreateSparseArray('Line', CreateDictionary('form', 'solid')),
-        'line.staff.vertical',                 CreateSparseArray('Line', CreateDictionary('form', 'solid', 'type', verticalLine, 'endid', 'PreciseMatch')),
+        'line.staff.nebenstimme',                   @Element('Line', @Attrs('startsym', 'N', 'endsym', 'angledown', 'lwidth', '0', 'endid', 'Previous')),
```
rettinghaus, line 52: In MEI 4 it's `width`:
```suggestion
        'line.staff.nebenstimme',                   @Element('Line', @Attrs('startsym', 'N', 'endsym', 'angledown', 'width', '0', 'endid', 'Previous')),
```
If you think this is the right solution, than it has to be also present on `line.staff.hauptstimme`.

rettinghaus (changes requested):
Looks mostly good. Some small comments from my side.

rettinghaus (approved):

Issue #74, opened 2017-03-12
Problems with beamed tuplets


rettinghaus:
in some cases `<tupletSpan>` elements are tripled. All carry the same `@startid` but only one of them a proper `@endid`. 

```xml
<tupletSpan xml:id="m-41" num="3" num.format="count" numbase="2" startid="#m-39"/>
<tupletSpan xml:id="m-43" num="3" num.format="count" numbase="2" startid="#m-39"/>
<tupletSpan xml:id="m-45" endid="#m-44" num="3" num.format="count" numbase="2" startid="#m-39"/>
<tupletSpan xml:id="m-51" num="3" num.format="count" numbase="2" startid="#m-49"/>
<tupletSpan xml:id="m-53" num="3" num.format="count" numbase="2" startid="#m-49"/>
<tupletSpan xml:id="m-55" endid="#m-54" num="3" num.format="count" numbase="2" startid="#m-49"/>
```

see:
https://github.com/DDMAL/mei-test-set/blob/gh-pages/MEI-test-set/MEI/meter/beaming-over-beats.mei

https://github.com/DDMAL/mei-test-set/blob/gh-pages/MEI-test-set/sibelius-files/meter/beaming-over-beats.sib

th-we:
There are more problems that are probably related. See the following conversion:

![grafik](https://user-images.githubusercontent.com/1147152/63576377-87b5d200-c57b-11e9-88a9-cbad7fd1831a.png)


```xml
<measure xml:id="m-34" label="1" n="1">
    <staff xml:id="m-36" n="1">
        <layer xml:id="m-37" n="1">
            <tuplet xml:id="m-40" bracket.visible="false" endid="#m-42" num="3" num.format="count" numbase="2" startid="#m-38">
                <beam xml:id="m-39">
                    <note xml:id="m-38" dur="16" dur.ges="43p" oct="5" pname="d" pnum="74" stem.dir="down" tstamp.ges="0s"/>
                    <note xml:id="m-41" dur="16" dur.ges="43p" oct="5" pname="c" pnum="72" stem.dir="down" tstamp.ges="0.098s"/>
                    <note xml:id="m-42" breaksec="1" dur="16" dur.ges="43p" oct="5" pname="d" pnum="74" stem.dir="down" tstamp.ges="0.199s"/>
                    <note xml:id="m-43" dur="16" dur.ges="64p" oct="5" pname="c" pnum="72" stem.dir="down" tstamp.ges="0.3s"/>
[...]

PR #229, opened 2025-05-09
Rewrite how layer containers (layer, beam, tuplet) are created

th-we:
Fixes long standing #74, the highest priority issue right now.

Previously, layer containers (layer, tuplets, beams) and "content" (notes, rests, chords, clefs) were all generated in one big step. Logic for how tuplets and beams were nested became scattered to different places and complex to follow, debug and fix.

This pull request separates the generation of layers, beams and tuplets from populating them with content.  Another change is that clefs are now encoded in all affected layers, with a `@sameas` pointing to the "primary" clef.

The last commits move some code out of `ExportGenerators.mss` because this is the largest file in the repo and is ever growing. For getting an overview of the changes in the logic, it might therefore be better to look at https://github.com/music-encoding/sibmei/compare/2c4a9b649d0098a23e68e0ee31f7702fb5ab0daa...8423af82accae9b60a42c135089df0372c84f882, where the changed code is still at its original place and you can see a better diff.

We are expecting a commission where we will render export results with Verovio and do detailed proofreading against the original Sibelius renderings.


th-we:
Rebased the branch on top of develop for easier testing without dependency on Gulp.

rettinghaus:
@th-we please do another rebase to resolve the conflicts.

th-we:
> During testing I had an error in `BuildNoteRestParentsByVoiceAndPosition` with the following piece:
> 
> [Liszt_-_S.136_No.10_-_Liszt__JUNPIN.mxl.zip](https://github.com/user-attachments/files/20968913/Liszt_-_S.136_No.10_-_Liszt__JUNPIN.mxl.zip)

Weird case! Sibelius' MusicXML import produces something weird in one bar.  A half note inside an eighth tuplet, and 3 more empty tuplets. I added a check to prevent a crash.

Will rebase later.

th-we:
Improved the last commit to prevent the crash and rebased to resolve conflicts.

PR #230, opened 2025-05-09
Rewrite generation of `<staffGrp>` hierarchy

th-we:
This pull request does the following things:

* Fix staff labels and support instrument labels.  Previously, Sibelius' `InstrumentName`/`ShortInstrumentName` values were used for staff labels, but they should be used for instrument labels. The `StaffName` values are now used for staff labels.
* Supports formatted instrument/staff labels
* Supports `@bar.thru`
* Avoids unnecessary `<staffGrp>` nesting. (For example a piano score would previously have two nested `<staffGrp>`s. The outer one would be there by default, the inner one would be created because there is a brace.)

Previously, only brackets and braces affected how staves were grouped.  In addition to brackets/braces, this pull request takes into account barline groups and multi-stave instruments.  For example, organ staves are now grouped into one staff group, even if there is a pedal staff that does not join a brace with the manual staves.


th-we:
I rebased the branch on top of develop to make review easier because Gulp is not required any more. I also added one small commit to export (relative) staff size.

Issue #206, opened 2023-02-13
Deprecate tido/plgToMSS and gulp


ahankinson:
Having recently got my installation back up and running, I'm reminded again of just how fragile JavaScript dependencies can be. I was only able to get things up and running by adding a `graceful-fs` override due to incompatibilities with gulp 4 and npm. So I'm thinking we should replace them with our own toolchain. This took about an hour and a half of Googling.

(Note: I'm not suggesting we deprecate the stuff used for testing, but this seems to be fairly well isolated.)

Having a look at the things we do use in the tido/plgToMSS repo, I think it's not anything that can't be replaced pretty easily. AFAIK we only use `buildPlg` and `deployPlg`. 

So my questions and comments are:

 - What language? Python is obvious, but might require some setup / Windows stuff? Bash is probably the most portable, but also a bit of a pain to write and maintain.
 - Replacing `plgconfig.js` -- I'm imagining the same thing but using YAML.
 - I'd like to keep having the recompile-on-change that the `develop` gulp command has, so that might have a bearing on what language we use.

This would be post-release.

craigsapp:
> Bash is probably the most portable, but also a bit of a pain to write and maintain.

PERL would be better than Bash since it is just as ubiquitous and less of a pain to write and easy to maintain. (in particular easier to maintain than Python code).

See:
https://github.com/rism-digital/verovio/blob/develop/emscripten/buildToolkit

Original bash script:
https://github.com/rism-digital/verovio/blob/5c220491196aa428f99ea9118bd027906de52c03/emscripten/build.sh

where I converted the original bash script for the emscripten compile command for verovio to PERL.

ahankinson:
Perl is a hard no from me.

craigsapp:
> Perl is a hard no from me.

Because you do not know it?   I am currently teaching a PERL class :-)

I could write a PERL script based on some current script as long as the source language is not too opaque.   As I do not use Sibelius (and by extenion sibmei), so I should not be a maintainer, of course.

ahankinson:
Given that the point of the exercise is to make the tools more maintainable, not less, then I'm not sure that writing it in a language that (I think) nobody here really uses is a good move.

craigsapp:
If you propose to write a script in bash but do not consider PERL, then I am quite curious as to why not?  You have not directly answered question, but it seems that it is due to lack of knowledge of programming in PERL.

https://xkcd.com/519

craigsapp:
[...]

PR #232, opened 2025-05-16
run npm audit

rettinghaus:
There were several vulnerabilities reported, so I ran `npm audit fix`. Seems to work fine for me, but please test locally before merging.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,12 +12,149 @@
         "ansi-colors": "^2.0.5",
         "fancy-log": "^1.3.2",
         "fontoxpath": "^2.12.1",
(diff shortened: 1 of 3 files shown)


th-we:
I tried running `npx gulp develp`, `npm test` and `npm run analyze`. Those worked as expected for me. However, the `*.sib` files get corrupted when they are copied to the `build` folder. Their size substantially increases. Is this the same on Mac?

th-we:
This is a good point of addressing #206. I'll take care of at least eliminating the gulp dependency.

th-we:
I created pull request #233 on top of this pull request to fix the issues.

PR #233, opened 2025-05-19
Eliminate dependencies on Tido compiler and gulp

### LINKED ISSUE #206: Deprecate tido/plgToMSS and gulp

th-we:
Fixes #206.

With full control over the compiler, I took the chance to hopefully improve the development experience:

* I removed the Testsibmei4 plugin. Instead, I now compile two versions of Sibmei: A release build and a develop build. The develop build is identical to the release build, except it has the tests directly compiled into it.  This is a tried and tested approach that I use for all our Sibelius plugins. It has the benefit that one does not have to reload both plugins. Reloading is a huge hassle by itself, and it gets even more of a pain when working on tests, having to remember which of the two plugins have been edited since the last reload. Tests can still be run using the minimal Test runner plugin, which hardly ever has to be changed and reloaded.
* I tried to remove all references to the version of sibmei ("sibmei**4**", MEI version etc.) from the code so that we can upgrade to the next major version without having to change that information in quite a few places, and without having to rename a number of files that have the major version in their name. The compiler now compiles global variables with that information into the plugins as global variables and adds a prefix to file names.

I simplified the README, removing the warnings, and for the sake of brevity and simplicity, I don't references both npm and yarn any more and just give the npm instructions.


ahankinson:
- Personally I prefer yarn over npm. Not a big deal, though.
- Would it make sense to fork sib-test to music-encoding if this will continue to be a dependency?? I'm not sure how much longer the tido repositories will be up.
- otherwise, it looks good.

th-we:
Alas, sib-test has no license attached. I have a fork [here](https://github.com/notengrafik/sib-test) in case the Tido repos disappear.

As npm is the default that is installed with node, there is one step less for any new developers.  Anyone who prefers yarn can of course use it. I updated the README to mention that git and node need to be installed.

rettinghaus (changes requested):
Looks good, just a tiny detail.

th-we:
> Looks good, just a tiny detail.

For some reason, I see no no requested changes.

README.md
```diff
@@ -12,90 +12,59 @@ If you're only interested in running the plugin in Sibelius, this is all you nee
...
 
-### Building using the NPM method
+You need to create a symbolic link from the Sibelius plugin folder to the build folder in the sibmei directory. The paths are system dependent.
 
-Previous versions of the plugin had you install packages externally, but in this version you can install everything locally within the development directory, including the Tido plgToMSS code (installed from GitHub) and gulp.
+Example for Mac:
 
-```shell
-npm install          // installs the packages listed in the sibmei package.json directory
+```bash
```
rettinghaus, line 31: Why change it to `bash`? If there are no specific bash commands I'd prefer sticking with the more generic `shell`.

th-we: Agreed. If that's all, can you merge?

ahankinson:
Sorry!

ahankinson (approved):

PR #236, opened 2025-06-27
add schema url for validation

rettinghaus:
This PR adds a processing instruction to the generated MEI enabling validation against the published RNG, which makes manual testing much easier.

NB: This complicates working with customizations, but for official releases of sibmei this should be fine.

### DIFF
--- a/lib/libmei4.plg
+++ b/lib/libmei4.plg
@@ -2045,7 +2045,8 @@ GetTail "(element) {
 
     _exportMeiDocument "(meidoc) {
         xdecl = '<?xml version=' & Chr(34) & '1.0' & Chr(34) & ' encoding=' & Chr(34) & 'UTF-16' & Chr(34) & ' ?>';
-        meiout = xdecl & convertDictToXml(meidoc[0], Chr(10));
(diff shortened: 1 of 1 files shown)


lib/libmei4.plg
```diff
@@ -2045,7 +2045,8 @@ GetTail "(element) {
 
     _exportMeiDocument "(meidoc) {
         xdecl = '<?xml version=' & Chr(34) & '1.0' & Chr(34) & ' encoding=' & Chr(34) & 'UTF-16' & Chr(34) & ' ?>';
-        meiout = xdecl & convertDictToXml(meidoc[0], Chr(10));
+        schema = '\n<?xml-model href=' & Chr(34) & 'https://music-encoding.org/schema/4.0.1/mei-CMN.rng' & Chr(34) & ' type=' & Chr(34) & 'application/xml' & Chr(34) & ' schematypens=' & Chr(34) & 'http://relaxng.org/ns/structure/1.0' & Chr(34) & ' ?>';
```
th-we, line 2048: Can you replace `4.0.1` with the global variable `MeiVersion`?

th-we, line 2048: Any reason to not also add Schematron validation (`<?xml-model href="https://music-encoding.org/schema/4.0.1/mei-CMN.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>`)?

th-we (changes requested):

th-we:
> NB: This complicates working with customizations, but for official releases of sibmei this should be fine.

We can give the extension API control over if and what to write for schema validation. I'll open an issue for this.

lib/libmei4.plg
rettinghaus: Not really. I'll add it.

PR #237, opened 2025-06-27
fix stacking of formatted text

### LINKED ISSUE #234: Title nested in title
Latest development version produces encoding errors within `anchoredText`. The title element there is created twice resulting in an invalid `title` within `title`:

```xml
      <anchoredText xml:id="m-67">
          <title xml:id="m-68">
              <title xml:id="m-69">
                  <rend xml:id="m-70" fontweight="bold">[Title]</rend>
              </title>
          </title>
      </anchoredText>
```

rettinghaus:
This PR fixes the invalid nested `title` elements. The `FormattedText` we are adding is actually a complete xml chunk, so no need to create a separate element.

closes #234


th-we:
This change would remove the `<title>` markup entirely for title text that does not use the `\$Title\` syntax.  I attached an example that has both a title text that references the title from the score's metadata and one that uses explicit text.

[titles.zip](https://github.com/user-attachments/files/20956721/titles.zip)

I think this will have to be addressed here instead:

https://github.com/music-encoding/sibmei/blob/692134f5560392ab4599a0f85fbe34539ee2e18d/src/TextHandler.mss#L426C32-L466

Other substitutions might lead to unintende nesting as well.

th-we:
I wrote an alternative solution to the problem. #239

th-we:
I don't fully understand your comment yet, but I wasn't able to have a look at the Sibelius file you sent as we didn't yet upgrade to a version supporting the new 2025.2 file format. Can you export to 2024.10? Sorry for the inconvenience! We're always cautious to upgrade, and if newer versions don't offer features or fixes we intend to use, we may stay on a tried and tested version for a while.

ahankinson:
Some comments beyond the conversation about the "best" way to do this, based on @rettinghaus' XML example.

1) Do we need `<seg>` here? It seems it's only function is to hold the `@type` attribute, but there is `<anchoredText>` `@func` for this?
2) The `<lb>` is not a line break but a line begin. So it should probably go before the element, not after it.
3) Not sure why the elements are backwards? (The dedication comes before the subtitle?) And doesn't the anchored text need an anchor? There's only a `@ho` on one, and none on the other...

rettinghaus:
I picked up the `@func` attribute coming from @ahankinson.
With this addition this looks now like

Bildschirmfoto 2025-06-30 um 18 46 51

```xml
<anchoredText xml:id="m-179" func="title">
    <title xml:id="m-180">
        <rend xml:id="m-181" fontweight="bold">Best piece ever</rend>
    </title>
</anchoredText>
<anchoredText xml:id="m-182" func="subtitle">
    <title xml:id="m-183" type="subordinate">for Brassband</title>
</anchoredText>
<anchoredText xml:id="m-184" func="composer"><persName xml:id="m-185" role="Composer"><rend xml:id="m-186" fontweight="bold">Johnny Vegas</rend></persName>&#10;<rend xml:id="m-187" fontstyle="italic">Arrangiert von </rend><persName xml:id="m-188" role="Arranger"><rend xml:id="m-189" fontstyle="italic">Knut Knot</rend></persName></anchoredText>
<anchoredText xml:id="m-190" func="dedication">
    <seg xml:id="m-191" type="Dedication">To the best community</seg>
</anchoredText>
```

when using referenced entries from the metadata. 
If you chose custom text, the result is the following:

Bildschirmfoto 2025-06-30 um 18 56 25

```xml
<anchoredText xml:id="m-46" func="title">
    <rend xml:id="m-47" fontweight="bold">This is the title field</rend>
</anchoredText>
<anchoredText xml:id="m-48" func="subtitle">with a subtitle</anchoredText>
<anchoredText xml:id="m-49" func="composer">
    <rend xml:id="m-50" fontstyle="italic">Peter van Achterhagen</rend>
</anchoredText>
<anchoredText xml:id="m-51" func="dedication">Dedicated to the fields of art</anchoredText>
```

rettinghaus:
I would suggest to change the text substitution later when moving to MEI5.

@ahankinson The text bits are not exported in the vertical order on the page, but in the chronological sequence they have been added. 🙄

th-we:
> Yes, but why do you insist on having a title tag there? I think `func="title"` and `func="subtitle"` is clear enough here. And you have the title tag already several times in the header.

Agreed, `@func` would suffice.

rettinghaus:
@th-we Then please test, approve and merge.

th-we (changes requested):
Tests would fail. Please merge the pull request I made to your branch.

th-we:
Thanks for spotting and the fruitful discussion!

PR #241, opened 2025-06-28
check for non-zero velocity

### LINKED ISSUE #238: All notes muted
Currently we export MIDI velocity on every note, which is derived from `OriginalVelocity`. The problem is, that this is always set to zero, if you enter the music note by note. 

The ManuScript documentation says about that

> Note that the word “original” refers to the fact that this data is preserved from the original performance if the score was imported from a MIDI file or input via Flexi-time.

As it is impossible to enter a "zero-velocity" note we should add this attribute only for non-zero values.

rettinghaus:
closes #238


th-we (changes requested):
Can you add a comment in the code explaining why this check is necessary? Because conceptually, a velocity of 0 seems to make sense e.g. for muting cues, but that does not work in Sibelius. (Muting requires unchecking the "Play on pass" checkboxes in the inspector, but I don't know if it's possible to access those values.)

th-we:
I created a pull request, adding a test file [here](https://github.com/rettinghaus/sibmei/pull/11). Can you merge that?

PR #243, opened 2025-07-14
change technique text indicator from label to type

rettinghaus:
Small change to align text output to `dir` with other methods where we prefer `@type` for indicating the text type.


th-we (approved):
Makes sense, works well.

PR #247, opened 2025-07-14
Fix crashes caused by invalid Sibelius.ActiveScore

### LINKED ISSUE #246: Test runner leads to Sibelius crash
On macOS with Sibelius 2025.4 Sibelius crashes when trying to use the `Test runner` plug-in. 
It seems to be connected with calling the `ActiveScore` method. I tried to fix one, but then the crash happens at the next instance.

Help would be appreciated.

th-we:
@rettinghaus, does this fix #246?

I also streamlined the functions a bit.  Intermediate assignments are not necessary.  I believe those assignments were originally used because Sibelius can indeed crash sometimes if values are not assigned to variables, but this basically only happens for TreeNodes, which frequently get garbage collected prematurely if not assigned to persistent variables.

An interesting observation I'd like to share:  This function:

```javascript
function ConvertOffsetsToMillimeters (score, offset) {
    return (score.StaffHeight / 128.0 * offset) & 'mm';
}  //$end
```
... has about the same performance as the old function using `Sibelius.ActiveScore`, but both are more than 100x slower than this function:
```javascript
function ConvertOffsetsToMillimeters (staffHeight, offset) {
    return (staffHeight / 128.0 * offset) & 'mm';
}  //$end
```
... or this function:
```javascript
function ConvertOffsetsToMillimeters (this, offset) {
    return (this.staffHeight / 128.0 * offset) & 'mm';
}  //$end
```
So passing around or retreiving the score object seems to have a tremendous overhead.  Even if the passed score is not used at all, this causes the 100x slowdown, like here:
```javascript
function ConvertOffsetsToMillimeters (staffHeight, score, offset) {
    return (staffHeight / 128.0 * offset) & 'mm';
}  //$end
```
Other objects don't have such an impact. E.g. passing the SystemStaff only has a marginal impact:
```javascript
function ConvertOffsetsToMillimeters (staffHeight, systemStaff, offset) {
    return (staffHeight / 128.0 * offset) & 'mm';
}  //$end
```
If however the system staff is used to access the score, the 100x slowdown is there again:
```javascript
function ConvertOffsetsToMillimeters (systemStaff, offset) {
    return (systemStaff.ParentScore.StaffHeight / 128.0 * offset) & 'mm';
}  //$end
```
It might therefore be a good idea to cache some things. This could be done like:
```javascript
Self._property:StaffHeight = score.StaffHeight;
```
Opinions welcome.

### DIFF
--- a/src/Run.mss
+++ b/src/Run.mss
@@ -16,12 +16,9 @@ function Run() {
 }  //$end
 
 
-function GetExportFileName () {
-    // get the active score object
-    activeScore = Sibelius.ActiveScore;
-
-    if (Sibelius.FileExists(activeScore.FileName)) {
-        scoreFile = Sibelius.GetFile(activeScore.FileName);
+function GetExportFileName (score) {
+    if (Sibelius.FileExists(score.FileName)) {
+        scoreFile = Sibelius.GetFile(score.FileName);
         activeFileName = scoreFile.NameNoPath & '.mei';
         activePath = scoreFile.Path;
     } else {
@@ -29,7 +26,6 @@ function GetExportFileName () {
         activePath = Sibelius.GetDocumentsFolder();
     }
 
-    // Ask to the file to be saved somewhere
     filename = Sibelius.SelectFileToSave('Save as...', activeFileName, activePath, 'mei', 'TEXT', 'Music Encoding Initiative');
 
     return filename;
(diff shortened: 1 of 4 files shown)


rettinghaus:
It seems to make sense to directly pass StaffHeight to `ConvertOffsetsToMillimeters` and `ConvertUnitsToPoints`, even if there was no gain in performance. 
If the caching can be introduced without much effort it sounds like a good idea.

src/Run.mss
```diff
@@ -115,6 +102,31 @@ function DoExport (score, filename) {
 }  //$end
 
 
+function SetGlobalsForScore (score) {
+    // Sets some globals with information about the currently procesed score.
```
rettinghaus, line 106:
```suggestion
    // Sets some globals with information about the currently processed score.
```

PR #249, opened 2025-07-15
fix: MEI 4 doesn't allow color on text elements

rettinghaus:
This is a small fix for textual control elements like `dir` and `dynam`, which are not member of `att.color` in MEI 4. 

Export for color of articulations has been added.

### DIFF
--- a/src/Utilities.mss
+++ b/src/Utilities.mss
@@ -432,11 +432,13 @@ function AddControlEventAttributes (bobj, element) {
         }
     }
 
-    if (bobj.Color != 0)
+    if (bobj.Type != 'Text')
     {
-        libmei.AddAttribute(element, 'color', ConvertColor(bobj));
+        if (bobj.Color != 0)
+        {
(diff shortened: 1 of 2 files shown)


ahankinson (approved):

src/Utilities.mss
```diff
@@ -432,12 +432,14 @@ function AddControlEventAttributes (bobj, element) {
         }
     }
 
-    if (bobj.Color != 0)
+    if (bobj.Type != 'Text')
```
th-we, line 435: I think that's not the check we need here. For example, `<harm>` is allowed to have color. This would prevent export of color e.g. for figured bass.

th-we (changes requested):

src/Utilities.mss
rettinghaus: In MEI 5 `color` is available for all control elements, see https://github.com/music-encoding/music-encoding/pull/1457.

PR #254, opened 2025-07-18
Improve system staff handling

th-we:
This fixes a problem with volta brackets. They did not work properly if they encompassed more than one measure.  This was previously untested, so I added a test file.  I could also eliminate a number of global variables and substantially reduce the code required for processing the system staff. While being at it, I also supported a few more barline types.

Instead of the unspecific `<ending type="open/closed">` I now write `@lendsym`, which is also rendered properly by Verovio.


src/Initialize.mss
```diff
@@ -48,6 +48,22 @@ function InitGlobals (extensions) {
...
+    BarlineTypeMap[SpecialBarlineDouble] = 'dbl';
+    BarlineTypeMap[SpecialBarlineFinal] = 'end';
+    BarlineTypeMap[SpecialBarlineInvisible] = 'invis';
+    BarlineTypeMap[SpecialBarlineNormal] = 'single';
+    BarlineTypeMap[SpecialBarlineDotted] = 'dotted';
+    BarlineTypeMap[SpecialBarlineThick] = 'heavy';
+    // no MEI equiv:
+    // BarlineTypeMap[SpecialBarlineBetweenStaves] = ' ';
+    // BarlineTypeMap[SpecialBarlineTick] = ' ';
+    // BarlineTypeMap[SpecialBarlineShort] = ' ';
```
rettinghaus, line 64: This would go into [`@bar.method`](https://music-encoding.org/guidelines/v4/data-types/data.barmethod.html).

rettinghaus:
Works great! Import with MuseScore:

grafik

src/Initialize.mss
th-we: Ah, right, I changed how the attributes are added.

th-we:
> Thanks for extending the test! Tiny thing: You export correctly `@bar.place`, but named it in the annotation `@bar.pos`.

Maybe I should save files before commiting? 🙄

PR #255, opened 2025-07-18
fix: deactivate test for MEI 4

rettinghaus:
We have a failing test for text color after #249. This PR simply deactivates this test for.


th-we:
O.K., I decided to ignore it because we'll re-activate it shortly anyway. That's very thorough of you.

PR #258, opened 2025-07-21
remove outdated yarn.lock file

rettinghaus:


rettinghaus:
Ah, I had the impression we dropped yarn with #233. Sorry to bother.

Issue #142, opened 2020-05-25
Sibmei 4 release


th-we:
I think we discussed earlier that we should keep Sibmei major version numbers in synch with MEI versions for the future. So I suggest we skip Sibmei 3 and rename `sibmei2.plg` to `sibmei4.plg` (same for `Testsibmei2.plg`) for the next release.

@ahankinson @annplaksin @rettinghaus: Agreed?

th-we:
Created pull request here #143.

PR #261, opened 2025-08-01
Update to MEI 5

th-we:
This is another really big pull request.

* Instead of re-compiling libmei for MEI 5, I merged the libmei code into the main plugin.  XML generation with the new code is now noticeably faster.
* With the old libmei functions that were generated for each element, attempting to create an invalid element would crash the plugin. Now, all elements are created with `CreateElement()`, which also checks if the element is valid. Instead of crashing the plugin, it will give a meaningful error message.
* I included RNG validation in the tests. As xmllint for some reason is unbearably slow for the MEI schema (no matter if it's used natively from the command line or as an npm package), I used jing instead.  It has to be installed on the system and `jing` must be on the path. If it is not, the test is skipped. In that case, a warning and installation instructions are shown in the test output.
* Static analysis has been extended to check for elements created by means of `CreateElement()` in addition to attributes that are created by means of `AddAttribute()`. Those checks are now also integrated in the node test suite.
* The ManuScript test suite checks if templates use valid element nesting and only attributes allowed on the respective elements.

The extended checks helped a lot in upgrading to MEI 5, and I hope they also will be for upgrading to future MEI versions.


th-we:
I rebased the branch to resolve conflicts.

package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "sibmei4",
-  "version": "4.1.0",
+  "name": "sibmei",
+  "version": "5.0.0",
```
ahankinson, line 3: Maybe keep the plugin in sync with the MEI version? 

```suggestion
  "version": "5.1.0",
```

src/Initialize.mss
```diff
@@ -91,6 +88,37 @@ function InitGlobals (extensions) {
         return false;
     }
 
+    Self._property:EncodedChar = CreateSparseArray();
+    EncodedChar[34] = '&quot;';
```
ahankinson, line 92: This should probably go in the `Xml` module...

test/sib-test/TestExportGenerators.mss
```diff
@@ -35,21 +35,21 @@ function TestGenerateMEIHeader (assert, plugin) {
...
 
-    d = libmei.getDocument();
+    d = GetDocument();
     filePath = Self._property:tempDir & 'header.mei';
-    e = libmei.meiDocumentToFile(d, filePath);
+    e = MeiDocumentToFile(d, filePath);
     assert.OK(e, 'The file ' & filePath & ' was successfully generated');
 
-    libmei.destroy();
+    InitXml();
```
ahankinson, line 47: Could you say why you renamed this? It seems like `InitXml` should be called before doing something, but this is always called after, which was the original meaning. 

Maybe you don't like `destroy`, but maybe something like `Clear()`?

package.json
th-we: I thought about that, but I don't think that's a good idea because we would only be able to make a release with a sensible version number when there's a new MEI version. For example, we're planning another Sibmei 4 release although there's no new MEI version. I think it's a good idea to keep the major version number in sync with the MEI schema and use our own schema for the rest.

test/sib-test/TestExportGenerators.mss
ahankinson: How about ResetXml?

src/Initialize.mss
ahankinson: I mean in `Xml.msd`

th-we:
Rebased branch to resolve conflicts and added commit to re-activate thick barlines that were de-activated in the meanwhile on the develop branch for the 4.2.0 release.

package.json
rettinghaus: Does the update to MEI 5 even justify the jump to a new major version number? I think the last release had more substancial changes than this. If we want to move to version 5 the main point here is probably dropping libmei …

ahankinson: #142 

We discussed this earlier. Verovio also keeps its major version numbers in sync with MEI.

ahankinson (approved):

ahankinson:
Thanks!

th-we:
Thank you for your patience with this big pull request, @ahankinson  and @rettinghaus !

PR #263, opened 2025-08-06
Last important fixes for MEI 4 

rettinghaus:
Three last small fixes to make exported MEI files vaild:

* do not try to export negative time (leads to bad formatted isotime)
* do not color syl elements
* do not (yet) export single heavy bar lines


th-we:
Thank you, @rettinghaus!

PR #268, opened 2025-08-08
Remove libmei from extension api. Expose functions marked with "export"

th-we:
All the libmei functions have been moved to the Sibmei plugin, but the extension API still has a field `libmei`. So that extensions still work, I provisionally dropped in sibmei5 itself for the old libmei. This however exposes all the functions present in Sibmei, which is not ideal.

In this pull request, I'm suggesting a mechanism to expose only specific functions. As we're using JavaScript-ish function syntax, I used the `export` keyword before functions that should be exposed. The mechansim goes as follows:

* The compiler recognizes the keyword and duplicates the function, adding the prefix `ExtensionAPI_` and a dummy as first parameter (which is needed for calling it as a method from the API object)
* It also compiles a list of all exported functions and 
* During plugin initialization, all the functions in the list are exposed in the api object

Like this, we can granularly and very easily expose only those functions we consider necessary for extensions without risking to break extensions.

I'm making this a draft pull request because documentation needs to be updated as well. Before continuing, I'm interested in your opinions whether this is a sensible way to go.

### DIFF
--- a/Extensions.md
+++ b/Extensions.md
@@ -61,10 +61,7 @@ Extensions must only interact with Sibmei through the API dictionary that is pas
 
 ### API data and methods
 
-The API dictionary exposes the following object:
 
-* **`libmei`**: A reference to libmei that can be used to construct and
-   manipulate MEI elements. *This dictionary must not be modified.*
 
 The API dictionary exposes the following [methods for registering Handlers](ExportHandlers.md#creating-and-registering-handlers) that must only be used inside the `InitSibmeiExtension()` method:
 


th-we:
> To me it looks like complicating stuff and making it less readable.

You mean the ManuScript code or the compiler code?

A strong motivation for having the `export` keyword right at the function: Changing the interface or the effects of such a function would break compatibility with the extension API. The `export` tag would be a reminder of this. If we expose functions by referencing them elsewhere, we might not notice right away that we're causing trouble.

th-we:
Anything missing for a Sibmei5 release when this has been merged?

PR #270, opened 2025-08-19
docs: add test.sh for mac

rettinghaus:


rettinghaus:
Thanks for bringing this up. I've added a note on this.

PR #299, opened 2026-03-27
Last instrument must receive a label, too

th-we:
Fixes #298


rettinghaus (approved):
Thanks! Maybe worth a bugfix release 5.0.1.

PR #305, opened 2026-07-29
Add support for SystemSymbolItems

th-we:
To my surprise, we didn't support `SystemSymbolItems` yet. @rettinghaus @annplaksin Could one of you check if the encoding makes sense? These are the added symbols and their `Index` property:

image


th-we:
I'm sorry, this breaks some seemingly unrelated tests. I have to fix that first.

th-we:
O.K., it's fixed now. Handling of `SystemTextItem`s had to change as well.

rettinghaus (approved):
Looks good to me, but cannot test currently. I'll leave it for someone else to merge.

th-we:
I think there's currently no one else to test this, so I'll merge (I only corrected the copy-paste error in the test description).

th-we:
What I should have said earlier: Thanks for your prompt and ongoing support, @rettinghaus!

PR #308, opened 2026-08-04
Update `<appInfo>` generation

th-we:
* Omit OS information. `Sibelius.OSVersionString` is buggy and reports the wrong OS version (at least for Windows). I guess it's not of much relevance anyways.
 * Use a template instead of imperative generation for making the tree structure clearer.
 * Templates using global variables and providing explicit IDs required some updates to `MeiFactory()`.

I was also wondering whether we could abandon the "speaking" IDs (`@xml:id="sibelius"`, `"sibmei"` or plugin names). With our current ID generation, a risk of a clash with IDs `sibelius` and `sibmei` does not exist, and a clash with a plugin name is not very likely, but I don't see an advantage either.


ahankinson:
OS _Version_ might not be relevant, but it at least tells us whether something was exported from Windows or Mac. I would keep it.

rettinghaus:
I agree with @ahankinson. Keep the `Sibelius.OSVersionString`, even if the information is limited to these strings:

```
Windows 95
Windows 98
Windows ME
Windows NT 3.x
Windows NT 4
Windows 2000
Windows XP
Windows Vista
Windows 7
Mac OS X
Mac OS X Jaguar
Mac OS X Panther
Mac OS X Tiger
Mac OS X Leopard
Mac OS X Snow Leopard
Unknown system version
```

Instead remove the (speaking) xml:ids. I don't see the need for those.

ahankinson:
Gosh. Can you imagine running Sibelius 7 on Windows 95?

rettinghaus:
(I'm wondering if they even care to update this list.)

ahankinson:
> (I'm wondering if they even care ~to update this list.~)

Fixed that for you.

th-we:
> I agree with @ahankinson. Keep the `Sibelius.OSVersionString`, even if the information is limited to these strings:
> 
> ```
> Windows 95
> Windows 98
> Windows ME
> Windows NT 3.x
> Windows NT 4
> Windows 2000
> Windows XP
> Windows Vista
> Windows 7
> Mac OS X
> Mac OS X Jaguar
> Mac OS X Panther
> Mac OS X Tiger
> Mac OS X Leopard
> Mac OS X Snow Leopard
> Unknown system version
> ```
> 
> Instead remove the (speaking) xml:ids. I don't see the need for those.

The [2026.06](https://resources.avid.com/SupportFiles/Sibelius/2026.6/ManuScript_Language_Guide.pdf#page=119) docs give the following list:

```
Windows 10
Windows 11
macOS Mojave
macOS Catalina
macOS Sonoma
macOS Sequoia
Unknown system version
```

I still have 2025.07 installed, for which the docs have the same list, but it reports Windows 8. I have no idea if the version is any more accurate on Mac. At least the listed Sequoia isn't up to date.

My suggestion: I only report what's before the first space character, then wie have "Windows", "Mac"/"macOS" (whatever it actually reports) and "Unknown".

ahankinson:
Sounds good!

th-we:
O.K., with the additional commits, the `<appInfo>` element now looks like this:
```xml
<appInfo xml:id="m-6">
    <application xml:id="m-7" type="music-notation-software" version="20250700">
        <name xml:id="m-8" type="application">Sibelius</name>
        <name xml:id="m-9" type="operating-system">Windows</name>
    </application>
    <application xml:id="m-10" isodate="2026-8-4T16:37:25Z" label="sibmei5" type="export-plugin" version="5.0.1">
        <name xml:id="m-11">Sibelius to MEI 5.1 Exporter</name>
    </application>
    <application xml:id="m-12" label="sibmei5_extension_test" type="sibmei-extension" version="5.0.1">
        <name xml:id="m-13">Sibmei extension test</name>
    </application>
</appInfo>
```
I added some `@type` and `@label` attributes to make clearer what's what.

I also added the version of the used extension plugins, if the extension has a `PluginVersion` global variable. Don't be confused by the fact that in this case, the version of the extension plugin is identical to the Sibmei version. That's only because the test extension is effectively part of Sibmei. 

I moved `@isodate` from Sibelius' `<application>` element to Sibmei's because that's effectively the export date where Sibmei was in use, while the creation of the Sibelius file might have been at any point in the past.

rettinghaus (approved):

th-we:
Thanks to both of you for your help!

PR #310, opened 2026-08-15
Rework NoteStyle export

th-we:
Previously, `NoteStyle` was converted to a `@head.shape` attribute (sometimes using attribute values that were neither in the specs nor related to an auth), but NoteStyles can influence more attributes. This PR also implements all built-in note styles and enables extensions to define how to export custom note styles.

Where possible, the head shapes are described by means of [`data.HEADSHAPE.list`](https://music-encoding.org/guidelines/v5/data-types/data.HEADSHAPE.list.html), otherwise SMuFL codepoints are used.

The templates are also checked against the schema to facilitate upgrading to newer MEI versions.  (I noticed this was missing for clef templates, so I added this as well). The existing Node based tests were replaced by XPath based tests embedded in the Sibelius test file.

Head encoding in MEI needs improvement (https://github.com/music-encoding/music-encoding/issues/865, https://github.com/music-encoding/music-encoding/issues/1153). This is an attempt at finding workable solutions based on the current specs.


rettinghaus (commented):
@th-we Before we merge, could you please open a ticket for MEI (or add something to https://github.com/music-encoding/music-encoding/issues/865 or https://github.com/music-encoding/music-encoding/issues/1153) to make your observations more visible? Otherwise they will be easily forgotten.

th-we:
True. Alas, I've already spent way more time on this topic than planned and first need to get back on track with other tasks that have deadlines.

th-we:
There will be merge conflicts with #311. The `extensions.sib` file changed for both PRs. This is the file that combines both:

[extensions.zip](https://github.com/user-attachments/files/31551764/extensions.zip)

th-we:
I rebased on top of current develop.

th-we:
I rebased this, making use of #316.

th-we:
~~Last rebase wasn't actually sensible and conflict free, so I rebased again on top of develop.~~ Sorry, the evening is too late. I'll have to come back to this with a fresh mind.

th-we:
Sorry, this PR is actually OK, but issues occurred because I apparently did not properly run tests for #321. I dared to push the fix for those directly to develop (https://github.com/music-encoding/sibmei/commit/d2a0b8e2d1f02ca17364ef9ddc1d529796e67b15) so that I can rebase all other PRs on top of develop to make them actually work.

rettinghaus (approved):
Would still be good, if you'd report to [music-encoding](https://github.com/music-encoding/music-encoding/).

PR #311, opened 2026-08-27
Support blank pages at any point of the score

th-we:
Previously, only leading blank pages were exported as `<front>`. (Well, Sibelius calls them "blank" pages, but they can have text and graphics content.) This PR exports content and `<pb>` elements for blank pages occurring at any point of the score.

I encode leading blank pages the same way as all other blank pages, using a generic `<div>` and dropped the encoding as `<front>`. We can't tell if those blank pages are actually used for front matter. If a project wants to encode leading blank pages as `<front>` (or trailing blank pages as `<back>`), they should add a postprocessing step to wrap those elements accordingly.

Previously, all text on blank pages was simply dumped as `<div><p>Some text</p></div>` without any information what kind of text it was. Now, style specific templates are used to e.g. export `<head><title>Some text</title></head>`.

This PR also reduces the size of the rather long functions `GenerateMEIMusic()` and `GenerateMeasure()` (but each still has 100+ lines) and eliminates a few global variables.

Tests were updated accordingly and the JavaScript based test for text was replaced with XPath based tests embedded in the Sibelius test file.


th-we:
There will be merge conflicts with #310. The `extensions.sib` file changed for both PRs. This is the file that combines both:

[extensions.zip](https://github.com/user-attachments/files/31551764/extensions.zip)

rettinghaus (approved):

PR #312, opened 2026-08-27
Add Sibmei version to test extension titles

th-we:
We had the situation that extension test plugins (e.g. `sibmei4_extension_test.plg` and `sibmei5_extension_test.plg`) registered the same plugin title. This leads to multiple indistinguishable entries in the extension choice dialog.
image
This PR fixes this. The new extensions have the version in their name.
image

### DIFF
--- a/test/extension_specific_schema.plg
+++ b/test/extension_specific_schema.plg
@@ -3,7 +3,7 @@
   CustomSchemaLocation "https://example.com/schema/custom.rng"
 
   Initialize "() {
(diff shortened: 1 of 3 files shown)


rettinghaus (approved):
Straight forward. I'm wondering if we can mark development builds as well, like  "Sibmei dev extension tests".

th-we:
Test extensions are only included in the development build. They're only there for testing purposes, not to be used by actual users.

rettinghaus:
Yes, good morning. Should activate my brain before commenting.

PR #313, opened 2026-09-01
Add `@tstamp.real` to all control events

th-we:
This is useful for playback oriented applications (when does the dynamic or tempo change occur?) and it's an easier alternative to `@tstamp` when analyzing the music for the order of events and alignment of elements.

I found that `ConvertTimeStamp()` was buggy and returned `00:60:00` for one hour. I fixed that and also made it about 15-20 times faster.

I added it in the `if` block that tests whether `@tstamp` should be written because `@tstamp.real` is available at all the elements that support `@tstamp` (plus some more).

### DIFF
--- a/src/ExportConverters.mss
+++ b/src/ExportConverters.mss
@@ -601,37 +601,24 @@ function ConvertTimeStamp (time) {
         return ' ';
     }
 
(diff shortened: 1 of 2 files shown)


rettinghaus (approved):
Thank you.

PR #315, opened 2026-09-14
Export more styles: Ritardando lines and some system texts

th-we:


src/LineHandler.mss
```diff
@@ -139,7 +139,35 @@ function InitLineHandlers () {
         'line.staff.hairpin.diminuendo.bracketed',  @Element('hairpin', @Attrs('form', 'dim')),
         'line.staff.hairpin.diminuendo.dashed',     @Element('hairpin', @Attrs('form', 'dim', 'lform', 'dashed')),
         'line.staff.hairpin.diminuendo.dotted',     @Element('hairpin', @Attrs('form', 'dim', 'lform', 'dotted')),
-        'line.staff.hairpin.diminuendo.tosilence',  @Element('hairpin', @Attrs('form', 'dim', 'niente', 'true'))
+        'line.staff.hairpin.diminuendo.tosilence',  @Element('hairpin', @Attrs('form', 'dim', 'niente', 'true')),
+
+        // Type = 'RitardLine'
+        'line.system.tempo.accel',                 @Element('tempo', @Attrs('extender', 'true', 'lform', 'dashed'), @Element('rend', @Attrs('fontweight', 'bold'), 'accel.')),
```
rettinghaus, line 145: Do you insist on having a `rend` everywhere for saying that it is bold? What happens if you change the style?

th-we: Sibelius has different variants of these "lines"/texts. Some are italic, some are bold, and that's reflected with `<rend>`.

Can you explain what precisely you mean by "if you change the style"?  If you switch say from one of the italic styles to one of the bold styles, then that's reflected by the `<rend>` elements.  When the look of the style is modified by the user and the export should reflect that, that's what extension plugins are for.  A template can be registered that overrides Sibmei's standard template.

th-we: Here's a screenshot of the test file to illustrate the different built-in styles:

image

rettinghaus: Everythiongs fine. Thanks for the explanation.

PR #316, opened 2026-09-14
Map SMuFL names to SmuflChar and SmuflHex values using global variables

th-we:
This is a PR that two other pull requests depend upon. It is intended to avoid magic numbers when working with SMuFL codepoints.


rettinghaus (approved):
lgtm

th-we:
Thanks! Any opinions on the TODO comment I had in there?  I.e. using the official SMuFL JSON to extract the names and codepoints?  We could could either

* keep maintaining SMuFL mappings like I did in this PR,
* copy SMuFL's JSON into the repo, 
* automatically fetch the JSON from https://smufl.formats.music/metadata/glyphnames.json (similar to how we currently fetch the MEI schemas for testing), 
* or add a git submodule (which would simplify pinning the version)

rettinghaus:
I'm in favor of the two latter. But I'm not sure how this file and the functionality then would end up in a release. Could you please explain?

th-we:
I would do it similarly to how we add information about the MEI schema to check for invalid element names. The compiler parses the schema and generates a global variable that looks like this:

```
LegalElements {
   "arpeg" {
      attributes {"xml:id" "xml:base" "label" "copyof" "corresp" "follows" "next" "precedes" "prev" "sameas" "synch" "n" "resp" "class" "type" "facs" "when" "layer" "part" "partstaff" "plist" "staff" "evaluate" "tstamp" "tstamp.ges" "tstamp.real" "startid" "order" "altsym" "color" "enclose" "glyph.auth" "glyph.uri" "glyph.name" "glyph.num" "lform" "lwidth" "lsegs" "fontfam" "fontname" "fontsize" "fontstyle" "fontweight" "letterspacing" "lineheight" "ho" "to" "vo" "x" "y" "arrow" "arrow.shape" "arrow.size" "arrow.color" "arrow.fillcolor"}
      children {}
    }
   "attacca" {
      attributes {"xml:id" "xml:base" "label" "copyof" "corresp" "follows" "next" "precedes" "prev" "sameas" "synch" "n" "resp" "class" "type" "facs" "xml:lang" "translit" "tstamp2.ges" "tstamp2.real" "when" "layer" "part" "partstaff" "plist" "staff" "evaluate" "tstamp" "tstamp.ges" "tstamp.real" "startid" "endid" "tstamp2" "target" "color" "lform" "lwidth" "lsegs" "lendsym" "lendsym.size" "lstartsym" "lstartsym.size" "extender" "place" "vgrp" "ho" "to" "vo" "startho" "endho" "startto" "endto" "x" "y"}
      children {"fig" "watermark" "catchwords" "heraldry" "locus" "locusGrp" "secFolio" "signatures" "stamp" "bloc" "country" "district" "geogFeat" "region" "settlement" "postBox" "postCode" "street" "periodName" "styleName" "ptr" "ref" "address" "annot" "bibl" "biblStruct" "date" "depth" "dim" "height" "width" "abbr" "expan" "identifier" "lb" "num" "corpName" "persName" "geogName" "repository" "name" "relation" "relationList" "rend" "stack" "title" "dedicatee" "dimensions" "extent" "symbol" "term" "q" "seg" "anchoredText" "curve" "line" "choice" "subst" "add" "corr" "damage" "del" "gap" "handShift" "orig" "reg" "restore" "sic" "supplied" "unclear"}
    }
    ...
```

Similarly, the compiler could parse the SMuFL JSON and generate a variable like this:
```
SmuflCodepoint {
    266D "csymAccidentalFlat"
    266E "csymAccidentalNatural"
    266F "csymAccidentalSharp"
    E047 "segno"
    E048 "coda"
    E082 "timeSig2"
    E084 "timeSig4"
    ...
```
`SmuflCodepoints` would then replace the `reverseSmuflMap` variable in `InitSmuflMaps()`.

ahankinson:
Do you need every symbol mapped if you’re only checking validity? Why not just check ranges?

th-we:
Only `LegalElements` is used for validation. `SmuflCodepoints` would be used for lookup of codepoints that are written to MEI, like here:

https://github.com/notengrafik/sibmei/blob/b2aece00039db66199566cac1885449ea4fbef15/src/Utilities.mss#L726-L808

Compiling all the codepoints into the plugin would of course mean a substantial increase in size. If we don't want to include the mappings in the plugin because of that but neither want to manually curate them, we could add some preprocessor functionality to the compiler and replace e.g. `@SmuflHex.csymAccidentalFlat` with `U-266D` at compile time based on the SMuFL JSON.

* Advantages: Smaller plugin size and very slightly faster execution because lookups are avoided.
* Disadvantage: The mappings can not be made available to extension plugins.

PR #317, opened 2026-09-14
Test with XQuery instead of XPath

th-we:
This rewrites the JavaScript code that evaluates the XPath tests embedded in Sibelius files and uses XQuery instead.  As XQuery is a superset of XPath, all existing tests remain functional. Additional XQuery functions `assert:equal()` and `assert:ok()` are now accessible to XPath/XQuery tests.


th-we:
The two reason for rewriting this were:
 
*  The old way of giving meaningful messages when a comparison failed (like `tempo = 'Andante'`) was pretty hacky and could break the test. The existing comparison tests will continue to work, but in case of a regression, they will now simply tell which XPath expression failed instead of telling us e.g. `expected: "Allegro", actual: "Andante"`. New tests should use `assert:equal()` for generating meaningful messages.
* I'll soon open a PR for guitar chord diagrams. Testing those requires resolving ID references which is hardly possible with XPath only and requires XQuery `let` assignments.
* When testing a node for multiple properties, like `staff[@n=3]/layer[@n=2]/note[1][@pname='a'][@oct=4]`, when the XPath test would fail, we wouldn't know the reason right away. With assertion functions, we can now get more information, like this: `staff[@n=3]/layer[@n=2]/note[1][assert:equal(@pname, 'a')][assert:equal(@oct, 4)]`, or:
    ```
    let $note := staff[@n=3]/layer[@n=2]/note[1]
    return (
        assert:equal($note/@pname, 'a', 'note name must be a'),
        assert:equal($note/@oct, 'a', 'expected octave 4')
    )
    ```

